PR #17, opened 2026-01-08
refactor: redundant dependencies remove

LukiLenkiewicz:
### Summary
I replaced pydantic-ai with pydantic-ai-slim. Thanks to this setting redundant dependencies are not installed.

### Changes
- replacing pydantic-ai with pydantic-ai-slim as a main package
- setting cli, web and sanbox dependencies as optional to avoid installation of unnecessary libraries


DEENUU1 (commented):
.

PR #81, opened 2026-04-19
fix: remove skill's main readme.md file from resources

ilayu-blip:
If the skill's repo contains a readme.md file, it will show as a resource of the skill, even though it should not.
The readme.md file at the root of the repo is for the repo itself.


ilayu-blip:
I was wrong :) I'm deleting this.

PR #96, opened 2026-05-08
fix: Esc interrupt, tool spinner lifecycle, and empty message cleanup

### LINKED ISSUE #93: Theres no /stop ?
### Description

so im testing, asking the agent just to go and create a directory, it cannot find the powershell? or do i have to configure it? the big problem is, when its on "execute" and its just doesnt want to load, i cant stop it.. i have to /quit, and even if i /save the session, when i /load it, im going back to that "execute" part where it just continuing its thing.. theres no /stop? .. im not going to /save everytime i want to execute something.. or that is what i should do?

Image

### Steps to Reproduce

explained above

### Expected Behaviour

Assistant can be stopped when execution went wrong

### Actual Behaviour

now almost 1000s

Image

### pydantic-deep Version

0.3.18

### Python Version

3.13.3

### Operating System

Windows 11

### Additional Context

_No response_

OchnikBartek:
## Summary
Four UX and reliability fixes for the TUI: Esc now interrupts a running agent, tool-call spinners always stop correctly, `/load` session replay no longer shows infinite spinners for previously interrupted calls, and the hints bar shows context-aware shortcuts while the agent is running.

## Added
- **Esc to interrupt**: `DeepApp.action_escape_key()` cancels the running agent task on Esc; focuses input when idle. Hint bar updates to `Esc  interrupt` while the agent runs and restores defaults on completion
- **Empty message cleanup**: if the agent is cancelled before producing any output, thinking, or tool calls, the empty assistant message bubble is removed from the UI

## Changed
- Agent task stored on `app._agent_task` so `action_escape_key` can cancel it
- `finally` drain in `_agent_stream_worker` distinguishes cancellation (`_run_cancelled` flag): only marks pending tool widgets "Interrupted" on actual `CancelledError`; on normal completion stops spinners without error state
- `complete_tool_call` is now idempotent — skips widgets already marked success/error, preventing the drain from overwriting a correctly completed tool call

## Fixed
- Tool-call spinner timer race condition in `ToolCallWidget.on_mount`: if `complete()` was called before mount (fast tools), the spinner interval is never started and the final state is rendered immediately
- `/load` session replay: orphaned tool calls (from previously interrupted sessions) are replayed as "Interrupted" instead of spinning forever; `part.args_as_dict()` used for correct label display
- Successful `execute` calls no longer shown as "Interrupted" after agent completes normally

## Removed
- `Binding("escape", "focus_input", ...)` from `ChatScreen.BINDINGS` — superseded by app-level `action_escape_key`

## Testing
Manually tested: `execute(sleep 5)` completes with green ✓; Esc mid-execution shows red ✗ "Interrupted"; `/load` of an interrupted session replays correctly; Esc when idle focuses the input.

## Related Issues
Closes #93

### DIFF
--- a/apps/cli/app.py
+++ b/apps/cli/app.py
@@ -45,6 +45,7 @@ class DeepApp(App):
 
     BINDINGS = [
         Binding("ctrl+c", "interrupt", "Interrupt", show=False),
+        Binding("escape", "escape_key", "Interrupt/Focus", show=False),
         Binding("ctrl+d", "quit", "Quit", show=False),
         Binding("f1", "show_help", "Help"),
         Binding("f2", "show_settings", "Settings"),
@@ -366,6 +367,18 @@ class DeepApp(App):
             self.notify("Agent interrupted", severity="warning")
         else:
             self.exit()
+
+    def action_escape_key(self) -> None:
+        """Handle Esc — interrupt running agent, or focus input if idle."""
+        if self._agent_task and not self._agent_task.done():
+            self._agent_task.cancel()
+            self.is_streaming = False
+            self.notify("Agent interrupted", severity="warning")
+        else:
+            from apps.cli.widgets.input_area import InputArea
+
+            with contextlib.suppress(NoMatches, Exception):
+                self.screen.query_one(InputArea).focus_input()
 
     def action_show_help(self) -> None:
         self.handle_command("/help")
--- a/apps/cli/app.py
+++ b/apps/cli/app.py
@@ -63,6 +63,8 @@ class DeepApp(App):
     total_cost: reactive[float] = reactive(0.0)
     current_cost: reactive[float] = reactive(0.0)
 
+    _agent_task: asyncio.Task[None] | None = None
+
     def __init__(
         self,
         agent: Agent[Any, str] | None = None,
@@ -83,7 +85,6 @@ class DeepApp(App):
         self._branch = _detect_git_branch(str(working_dir))
         self.message_history: list[ModelMessage] = message_history or []
         self.last_response: str = ""
-        self._agent_task: asyncio.Task[Any] | None = None
         self._startup_error = startup_error
 
         # Register custom themes
@@ -363,7 +364,6 @@ class DeepApp(App):
         """Handle Ctrl+C — cancel running agent or exit."""
         if self._agent_task and not self._agent_task.done():
             self._agent_task.cancel()
-            self.is_streaming = False
             self.notify("Agent interrupted", severity="warning")
         else:
             self.exit()
@@ -372,12 +372,11 @@ class DeepApp(App):
         """Handle Esc — interrupt running agent, or focus input if idle."""
         if self._agent_task and not self._agent_task.done():
             self._agent_task.cancel()
-            self.is_streaming = False
             self.notify("Agent interrupted", severity="warning")
         else:
             from apps.cli.widgets.input_area import InputArea
 
-            with contextlib.suppress(NoMatches, Exception):
+            with contextlib.suppress(NoMatches):
                 self.screen.query_one(InputArea).focus_input()
 
     def action_show_help(self) -> None:
--- a/apps/cli/commands.py
+++ b/apps/cli/commands.py
@@ -306,6 +306,14 @@ async def dispatch_command(app: DeepApp, command: str) -> None:  # noqa: C901
                 msg_list = app.screen.query_one(MessageList)
                 msg_list.clear_messages()
 
+                completed_call_ids: set[str] = set()
+                for msg in history:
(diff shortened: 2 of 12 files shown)


apps/cli/app.py
```diff
@@ -367,6 +368,18 @@ def action_interrupt(self) -> None:
         else:
             self.exit()
 
+    def action_escape_key(self) -> None:
+        """Handle Esc — interrupt running agent, or focus input if idle."""
+        if self._agent_task and not self._agent_task.done():
+            self._agent_task.cancel()
```
DEENUU1, line 374: asyncio.Task.cancel() schedules cancellation — the worker's finally block hasn't run yet when this line executes. If is_streaming gates the ability to submit
a new message, the user can send a new prompt while the old task is still in the middle of cleanup (saving session, draining spinners). Move self.is_streaming= False into the finally block of _agent_stream_worker, alongside header.is_streaming = False.

apps/cli/screens/chat.py
```diff
@@ -440,9 +441,14 @@ def _run_agent(self, text: str) -> None:
         app.last_response = ""  # type: ignore
         assistant = msg_list.begin_assistant_message()
 
+        with contextlib.suppress(Exception):
+            self.query_one(HintsBar).update("[dim]Esc[/dim] to interrupt")
+
         task = asyncio.create_task(self._agent_stream_worker(text, assistant, msg_list, header))
+        app._agent_task = task  # type: ignore[attr-defined]
```
DEENUU1: Three # type: ignore[attr-defined] suppressions across the PR signal this should be a proper class attribute. Add it to DeepApp: _agent_task: asyncio.Task[None] | None = None

apps/cli/screens/chat.py
```diff
@@ -778,13 +790,35 @@ def _parse_args(raw: Any) -> dict[str, Any]:
...
+
+            for call_id, (_, start) in list(pending.items()):
+                elapsed = _time.monotonic() - start
+                if _run_cancelled:
+                    assistant.complete_tool_call(call_id, "Interrupted", elapsed, True)
+                else:
+                    assistant.complete_tool_call(call_id, "", elapsed, False)
+            pending.clear()
+
+            is_empty = (
```
DEENUU1: Five private attributes accessed directly on two different widget classes (_text, _thinking, _tool_widgets on AssistantMessage; _current_assistant on MessageList). Any internal rename breaks this silently. Add a public is_empty: bool property on AssistantMessage and a remove_last_if_empty() method on MessageList to encapsulate this.

apps/cli/screens/chat.py
```diff
@@ -778,13 +790,35 @@ def _parse_args(raw: Any) -> dict[str, Any]:
...
+
             # Always save session — even after errors or cancellation
             self._save_session()
             header.is_streaming = False
             header.is_thinking = False
             msg_list.end_assistant_message()
             with contextlib.suppress(Exception):
                 self.query_one(InputArea).focus_input()
+            with contextlib.suppress(Exception):
+                self.query_one(HintsBar).update(HintsBar._default_hints())
```
DEENUU1: _default_hints is a private method (underscore prefix), called externally in two places in this PR. Either rename it to default_hints() (public) or expose a reset() method on HintsBar that encapsulates this.

apps/cli/commands.py
```diff
@@ -306,6 +306,14 @@ async def _handle_load(session_id: str | None) -> None:  # noqa: C901
                 msg_list = app.screen.query_one(MessageList)
                 msg_list.clear_messages()
 
+                completed_call_ids: set[str] = set()
+                for msg in history:
+                    for part in msg.parts:
```
DEENUU1: Consistent with the pre-existing pattern in this file, but string-matching on part_kind is fragile compared to isinstance(part, ToolReturnPart). Since
pydantic-ai exports these types, the isinstance check would be safer and type-checker-visible. Not blocking given the existing pattern, but worth a follow-up cleanup.

apps/cli/app.py
```diff
@@ -367,6 +368,18 @@ def action_interrupt(self) -> None:
...
+            self.is_streaming = False
+            self.notify("Agent interrupted", severity="warning")
+        else:
+            from apps.cli.widgets.input_area import InputArea
+
+            with contextlib.suppress(NoMatches, Exception):
```
DEENUU1: Exception already catches NoMatches (it's a subclass). Listing NoMatches alongside Exception is dead code. More importantly, suppressing all Exception here hides any unexpected errors silently. Either use just NoMatches (which is the only expected failure), or if something else can throw, be explicit about it.

DEENUU1 (changes requested):

PR #97, opened 2026-05-08
fix: add skills parameter to create_deep_agent

### LINKED ISSUE #95: Examples/full_mapp skill loading error
### Description

Image

### Steps to Reproduce

1. git clone https://github.com/vstorm-co/pydantic-deepagents.git
2. uv sync
3. cd examples/full_app 
4. uvicorn app:app --reload --port 8080

### Expected Behaviour

.

### Actual Behaviour

.

### pydantic-deep Version

0.3.18

### Python Version

3.12

### Operating System

macos

### Additional Context

_No response_

OchnikBartek:
## Summary

`create_deep_agent()` did not accept a `skills` parameter for passing programmatic `Skill` instances directly. Callers using `skills=...` (as shown in `examples/full_app/app.py`) received `pydantic_ai.exceptions.UserError: Unknown keyword arguments: 'skills'` because the kwarg was forwarded to pydantic-ai's `Agent()` constructor, which doesn't know it.

## Changes

- Added `skills: list[Skill] | None = None` to all three signatures of `create_deep_agent` (both `@overload`s and the implementation)
- Imported `Skill` from `pydantic_deep.toolsets.skills` in `agent.py`
- Threaded `skills` into `SkillsToolset(skills=skills, ...)` so programmatic skill instances are registered
- Added 3 new tests to `TestAgentIntegration` in `tests/test_skills_extended.py`

## Testing

- [x] New tests added for any new functionality

## Related Issues

Fixes #95

### DIFF
--- a/pydantic_deep/agent.py
+++ b/pydantic_deep/agent.py
@@ -499,6 +499,15 @@ def create_deep_agent(  # noqa: C901
         result = await agent.run("Analyze this code", deps=deps)
         ```
     """
+    import warnings
+
+    if not include_skills and (skills or skill_directories):
+        warnings.warn(
+            "skills and skill_directories are ignored when include_skills=False",
+            UserWarning,
+            stacklevel=2,
+        )
+
     model = model or DEFAULT_MODEL
     backend = backend or StateBackend()
     interrupt_on = interrupt_on or {}
--- a/pydantic_deep/agent.py
+++ b/pydantic_deep/agent.py
@@ -23,7 +23,7 @@ from subagents_pydantic_ai import create_subagent_toolset, get_subagent_system_p
 
 from pydantic_deep.deps import DeepAgentDeps
 from pydantic_deep.prompts import BASE_PROMPT
-from pydantic_deep.toolsets.skills import SkillsToolset
+from pydantic_deep.toolsets.skills import Skill, SkillsToolset
 from pydantic_deep.toolsets.skills.backend import BackendSkillsDirectory
 from pydantic_deep.types import SubAgentConfig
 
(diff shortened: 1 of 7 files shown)


pydantic_deep/agent.py
```diff
@@ -694,6 +698,7 @@ def _default_deep_agent_factory(cfg: dict[str, Any]) -> Any:
 
         skills_toolset = SkillsToolset(
             id="deep-skills",
+            skills=skills,
```
DEENUU1, line 710: This block runs only when include_skills=True. If a caller passes skills=[my_skill] alongside include_skills=False, the list is silently dropped — no error, no warning. The user gets an agent that ignores their skills with no feedback.

tests/test_skills_extended.py
```diff
@@ -671,6 +671,42 @@ def test_create_agent_skills_none(self) -> None:
             )
         assert agent is not None
 
+    def test_create_agent_with_skills(self) -> None:
```
DEENUU1, line 674: The original bug (#95) was a UserError on creation — so a smoke test catches the regression. But none of the three new tests verify that the skills are actually accessible on the resulting SkillsToolset. A test that inspects the registered toolsets (or checks the system prompt contains the skill) would give real confidence.

pydantic_deep/agent.py
```diff
@@ -304,6 +307,7 @@ def create_deep_agent(  # noqa: C901
         subagents: Subagent configurations for the task tool.
         skill_directories: Directories to discover skills from.
             Accepts plain string paths or BackendSkillsDirectory instances.
+        skills: Skill instances to register directly.
```
DEENUU1, line 310: Missing: what happens when combined with skill_directories, and what happens when include_skills=False.

DEENUU1 (changes requested):

PR #119, opened 2026-05-24
fix: migrate from _read_bytes to read_bytes for pydantic-ai-backend 0.2.8

mcauthorn:
## Summary
Version:  0.3.21

`pydantic-ai-backend 0.2.8` renamed `BackendProtocol._read_bytes(path)` → `read_bytes(path)` ([changelog](https://github.com/vstorm-co/pydantic-ai-backend/blob/HEAD/CHANGELOG.md#028---2026-05-24)), promoting the bytes-read entry point from private to public. 

Importantly, the CHANGELOG marks this as breaking, and no transitional alias was kept.

`pydantic-deep` still called the underscored name at five sites and declared its dependency as `pydantic-ai-backend>=0.2.7` (unbounded), so fresh resolutions pulled 0.2.8 transitively. Every toolset that touches the backend then failed at `get_instructions()` time with: 
`AttributeError: 'LocalBackend' object has no attribute '_read_bytes'`

This probably affects any non-trivial deepagent setup. 

## Changes
1. **Rename call sites** in the four toolsets that read bytes through the backend protocol:
   - `pydantic_deep/toolsets/memory.py` (`load_memory`)
   - `pydantic_deep/toolsets/liteparse.py` (PDF / file ingestion)
   - `pydantic_deep/toolsets/context.py` (`ContextToolset` preload)
   - `pydantic_deep/toolsets/skills/backend.py` (`BackendSkillResource` + `BackendSkillScript` load)
   Also updates two module-level docstrings in `skills/backend.py` that reference the protocol method by name.
2. **Update tests** that mocked or asserted on the same name:
   - `tests/test_memory.py`
   - `tests/test_liteparse.py`
   - `tests/test_skills_backend.py`
   - `tests/test_eviction.py`
3. **Bump dependency constraint** from `pydantic-ai-backend>=0.2.7` to `>=0.2.8` (both the main dep and the `sandbox` extra) so the declared minimum matches the protocol the code now targets.
## Relationship to PR #118
This supersedes #118 (Renovate auto-bump), which raises the constraint to `>=0.2.8` but doesn't update the code. 

I'm concered that merging #118 alone would force the broken combination on every installer; this PR makes the code and the constraint consistent.

## Test plan
- [x] Full suite passes: **1553 passed, 0 failed**
- [x] Directly affected modules pass: **227 passed** across `test_memory`, `test_liteparse`, `test_skills_backend`, `test_eviction`
- [x] No `_read_bytes` references remain in `pydantic_deep/` or `tests/`

### DIFF
--- a/pydantic_deep/toolsets/context.py
+++ b/pydantic_deep/toolsets/context.py
@@ -87,7 +87,7 @@ def load_context_files(
     """
     result: list[ContextFile] = []
     for path in paths:
-        raw = backend._read_bytes(path)
+        raw = backend.read_bytes(path)
         if not raw:
             continue
         content = raw.decode("utf-8", errors="replace")
@@ -115,7 +115,7 @@ def discover_context_files(
     found: list[str] = []
     for name in filenames:
         path = f"{search_path.rstrip('/')}/{name}"
-        raw = backend._read_bytes(path)
+        raw = backend.read_bytes(path)
         if raw:
             found.append(path)
     return found
--- a/pydantic_deep/toolsets/liteparse.py
+++ b/pydantic_deep/toolsets/liteparse.py
@@ -120,7 +120,7 @@ class LiteparseToolset(FunctionToolset[Any]):
             if not _HAS_LITEPARSE:
                 return _NOT_INSTALLED_MSG
             backend = ctx.deps.backend
-            file_bytes: bytes | None = backend._read_bytes(path)
+            file_bytes: bytes | None = backend.read_bytes(path)
             if not file_bytes:
                 return f"File not found: {path}"
             try:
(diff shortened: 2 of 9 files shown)


DEENUU1:
Thanks for catching this @mcauthorn — really nice debug, and the analysis in the description (resolution pulls 0.2.8 transitively, no deprecation alias kept) is spot on 🙌

Closing as superseded by **#118**, which is strict-superset: same rename, plus the parallel `summarization-pydantic-ai>=0.1.5` bump that Renovate batched together. **Your authorship is preserved in #118 via a `Co-authored-by` trailer on the rename commit**, so `git log` and GitHub will both credit you when it lands on `main`.

Will also mention you in the next CHANGELOG entry when 0.3.22 ships. Appreciated! 🚀

Issue #107, opened 2026-05-13
Stage 6: Autonomous merge (`JudgeAgent`)


OchnikBartek:
Part of #101.

## Goal

Replace manual-only resolution with autonomous picking by a cheap judge model. Default mode becomes `auto_with_fallback`: the judge picks, and if confidence is ≥ 0.80, auto-accept with a 2-sentence rationale; otherwise fall back to the manual picker with the judge's choice preselected. Also adds `auto` and `vote` modes.

This is where the feature crosses from "useful primitive" to *"git branch for cognition"* — users stop reviewing every fork manually.

## Architectural placement

```
pydantic_deep/
├── toolsets/
│   └── forking/
│       ├── judge.py                     # JudgeAgent, confidence scoring, modes
│       └── coordinator.py               # + resolve() extension for non-manual modes
└── types.py                             # + JudgeVerdict, ConfidenceSignals, BranchOutcome

apps/cli/
├── widgets/
│   └── merge_acceptance.py              # 2-sentence accept widget
└── modals/
    └── merge_picker.py                  # + judge preselect support
```

## Core design

The judge is a thin `Agent` wrapper with structured output. It receives **minimal** context — not the full per-branch history — to keep cost predictable and avoid reasoning over noise:

```python
class JudgeAgent:
    def __init__(self, model: str) -> None:
        self.agent = Agent(
            model,
            output_type=JudgeVerdict,
            system_prompt=JUDGE_SYSTEM_PROMPT,
        )

    async def evaluate(
        self,
        goal: str,                        # parent run's first user message
        diff_report: BranchDiffReport,    # from Stage 2
        outcomes: list[BranchOutcome],    # final assistant message + cost + turns + errors
    ) -> JudgeVerdict:
        prompt = _build_judge_prompt(goal, diff_report, outcomes)
        return (await self.agent.run(prompt)).output
```

The judge sees: the original goal, the structured diff report, and per-branch outcome summaries (final message + stats). It does **not** see each branch's full message history — keeps the prompt bounded.

### Confidence scoring

Three signals combined into a weighted score:

| Signal | Source | Weight |
|---|---|---|
| `quality_spread` | `1 - agreement_score` from `BranchDiffReport` (high = branches diverge meaningfully) | 0.4 |
| `test_pass_ratio` | passed tests / total tests across branch tool history (if any tests ran) | 0.4 |
[...]

Issue #106, opened 2026-05-13
Stage 5: External diff tool integration


OchnikBartek:
Part of #101.

## Goal

Materialize each branch's working tree to disk at `.pydantic-deep/forks/{fork_id}/branches/{label}/` (with parent snapshot at `.../parent/`), and invoke the user's external diff tool — `pycharm diff` (native 3-way) or `code --diff` (pairwise) — for visual review. Falls back gracefully to the TUI diff explorer when no editor is detected.

## Architectural placement

```
pydantic_deep/
└── toolsets/
    └── forking/
        ├── materializer.py              # disk flush, manifest, cleanup
        ├── editor.py                    # detection + subprocess invocation
        └── isolation.py                 # + hook materializer into BranchOverlay

apps/cli/
└── commands.py                          # + /fork open-diff <path>
```

## Core design

The materializer hooks into `BranchOverlay.write()` from Stage 1 — every write is mirrored to disk in real time, so editors auto-reload as branches progress:

```python
class ForkMaterializer:
    root: Path                            # .pydantic-deep/forks/{fork_id}
    fork_id: str

    async def snapshot_parent(self, paths: Iterable[str]) -> None: ...
    async def flush_change(self, branch_label: str, change: FileChange) -> None: ...
    def manifest_path(self) -> Path: ...      # JSON: branch_id ↔ label ↔ status
    def cleanup(self) -> None: ...            # on merge resolution or abort
```

Layout on disk:

```
.pydantic-deep/forks/fork_abc12/
├── manifest.json
├── parent/
│   └── src/auth/middleware.py
└── branches/
    ├── approach_a/
    │   └── src/auth/middleware.py
    ├── approach_b/
    │   └── src/auth/middleware.py
    └── approach_c/
        └── src/auth/middleware.py
```

Editor detection — priority order:

1. `PYDANTIC_DEEP_DIFFTOOL` env var (explicit override)
2. `pycharm` on `PATH` → native 3-way diff
3. `code` on `PATH` → pairwise (parent ↔ each branch)
4. `TUI fallback` → reuses the Stage 3 diff explorer

Invocation:

- PyCharm: single `subprocess.Popen(["pycharm", "diff", parent, a, b, c])`
- VS Code: N-1 `subprocess.Popen` calls, each pairing parent with one branch
- TUI fallback: opens the Stage 3 `MergePickerModal` in diff-explore mode

## CLI command

[...]

Issue #105, opened 2026-05-13
Stage 4: N branches + per-branch budget


OchnikBartek:
Part of #101.

## Goal

Remove the 2-branch cap from Stage 1. Enforce `budget_usd` per branch via `CostTracking.for_run()` — a branch exceeding its budget is terminated with status `budget_exhausted`, others continue running. Add an aggregate cost view across the fork and an optional fork-wide budget cap.

## Architectural placement

Pure extension of Stage 1 and Stage 3 — no new modules.

```
pydantic_deep/
├── toolsets/
│   └── forking/
│       └── coordinator.py               # + BudgetWatcher, fork_cost tool
└── types.py                             # + ForkCostSummary, BranchCost

apps/cli/
└── widgets/
    ├── fork_tabs.py                     # + per-branch cost in status badge
    └── side_panel.py                    # + aggregate cost in ForkBadgeWidget
```

## Core design

Each branch's runtime subscribes to its own `CostTracking.for_run()` instance (`for_run()` isolation already guarantees per-branch tracking — we just wire a callback):

```python
class BranchRuntime:
    cost_tracker: CostTracking
    budget_usd: Decimal | None

    async def _on_cost_update(self, info: CostInfo) -> None:
        if self.budget_usd and info.cumulative_usd >= self.budget_usd:
            await self.coordinator.terminate(
                self.id,
                reason=f"budget_exhausted ({info.cumulative_usd} >= {self.budget_usd})",
            )
```

`CostTracking` already emits `on_cost_update` callbacks (host apps consume them today); per-branch isolation comes for free from `for_run()`.

Rest:

- **`max_branches`** raised to `10` by default (configurable via `LiveForkCapability(max_branches=N)`).
- **`max_depth`** unlocked to `2` (was hard-pinned to `1` in Stage 1) — fork-of-fork now supported up to one level of nesting.
- **`fork_cost(fork_id)`** — new agent-facing tool returning `ForkCostSummary` with per-branch breakdown + aggregate.
- **`fork_run(..., aggregate_budget_usd=Decimal("5.00"))`** — optional sum-cap across all branches; when the aggregate is hit, the coordinator terminates all running branches with reason `aggregate_budget_exhausted`.
- **CLI status** — `ForkBadgeWidget` shows `[FORK: 3 branches · $1.24 / $5.00]` aggregate; tab status badges show per-branch spend.

## New types

Added to `pydantic_deep/types.py`:

[...]

Issue #104, opened 2026-05-13
Stage 3: CLI integration


OchnikBartek:
Part of #101.

## Goal

Surface forks in the TUI: `/fork` command + modal to spawn branches, per-branch tabs, per-branch steering via `MessageQueue` (`!a stop and focus on tests`), Esc to terminate the focused branch, `/merge` picker to resolve. Manual merge only — autonomous flow comes in Stage 6.

## Architectural placement

```
apps/cli/
├── forking.py                           # command handlers, coordinator bridge
├── widgets/
│   ├── branch_panel.py                  # per-branch streaming view
│   ├── fork_tabs.py                     # tab strip + status badges
│   ├── fork_overview.py                 # all-branches summary before merge
│   └── side_panel.py                    # + ForkBadgeWidget
├── modals/
│   ├── fork_picker.py                   # /fork input modal
│   └── merge_picker.py                  # /merge resolution modal
├── commands.py                          # + /fork, /merge dispatch
└── screens/
    └── chat.py                          # + !{branch_id} input routing
```

No changes to `pydantic_deep/` — purely a CLI consumer of Stage 1 and Stage 2.

## Core design

Reuses `MessageQueue` from #100 — `!{branch_id}` is a small extension of the existing `!` prefix routing. The CLI input router gains a regex check before falling through to the standard `MessageQueue.steer()` path:

```python
async def on_user_submitted(self, text: str) -> None:
    if active_fork := self.app.active_fork:
        if match := re.match(r"^!(\w+)\s+(.*)", text):
            branch_id, msg = match.groups()
            await active_fork.steer_branch(branch_id, msg)
            return
    # fall back to existing #100 behaviour
```

Rest is widgets and modals:

- **`BranchPanelWidget`** mirrors `ChatPanel` — same tool-call rendering, same streaming event handlers, just bound to a per-branch message stream. Most of it is composition over existing widgets.
- **`ForkTabsWidget`** is a tab strip with status badges per branch (`●` running, `✓` done, `✗` failed, `⊘` terminated). `Tab` cycles focus.
- **`ForkOverviewWidget`** is a "+" pseudo-tab showing all branches with summary stats — entry point for `/merge`.
- **`ForkBadgeWidget`** in the side panel (mirrors `QueuedWidget` from #100) shows `[FORK: 2 branches · $0.34]` while a fork is active, auto-hides on resolve.
[...]

PR #123, opened 2026-05-27
feat: live run forking

### LINKED ISSUE #101: Live Run Forking
Add a `LiveForkCapability` that lets an in-flight `agent.run()` split into N parallel branches sharing history up to the fork point, each diverging with its own steer message. A coordinator resolves the fork via one of four acceptance modes (`manual`, `auto`, `auto_with_fallback` default, `vote`). The winning branch's history is adopted as the parent run's continuation.

Natural next step of the live-control direction already set by `/remind` (#94), `MessageQueue` (#100), and Esc interrupt (#96). The agent becomes a forkable process — *git branch, but for cognition*.

## Why

When working with a long-running agent, exploring multiple solutions to the same problem in parallel is painful today. To try two implementation styles, hedge between fix hypotheses, or compare model outputs, you have to stop the agent, manually copy conversation history, start separate runs, and track them by hand. That loses the prompt cache, drops in-flight tool state, and quickly becomes unmanageable past two attempts.

Existing live-control primitives all treat the run as a single linear timeline. There is no way to *branch* an active run mid-flight — even though most of the underlying machinery (`CheckpointMiddleware`, `AgentTeam`, per-run isolation via `for_run()`, `clone_for_subagent`) is already in place.

Workflows this unlocks:
- Generating multiple implementation alternatives and comparing diffs
- Testing competing bug hypotheses in parallel, terminating the wrong ones
- Running the same task on different models to compare quality vs. cost
- Risk-hedging — conservative and aggressive variants side by side
- Cost calibration — same task at different budget caps

No competing framework offers live branching of a running agent. LangChain forks code, CrewAI has no fork primitive.

## How it would look

```python
from pydantic_deep import create_deep_agent, BranchSpec, BranchIsolation, MergeStrategy

agent = create_deep_agent(
    model="anthropic:claude-sonnet-4-6",
    forking=True,
)

# Mid-run, agent or user calls:
handle = await fork_run(
    branches=[
        BranchSpec(label="functional", steer="Implement using pure functions and immutable data."),
        BranchSpec(label="oop",        steer="Implement using classes with clear responsibilities."),
        BranchSpec(label="hybrid",     steer="Mix functional core with thin OO wrapper.",
                   model="anthropic:claude-haiku-4-5"),
    ],
    isolation=BranchIsolation(backend="copy", message_queue="isolated"),
    strategy=MergeStrategy(kind="auto_with_fallback", confidence_threshold=0.80),
)
```

CLI surface: `/fork` to spawn, tab navigation between branch views, per-branch steering via `!{branch_id} message` (extending the `!` prefix from #100), `/merge` to resolve.

## Acceptance modes

- **`manual`** — user compares branches in TUI diff explorer (or via an external diff tool like `pycharm diff` / `code --diff` if detected) and picks
- **`auto`** — cheap judge model picks autonomously, user sees only the result
- **`auto_with_fallback`** *(default)* — judge picks; if confidence ≥ 0.80, auto-accepts with 2-sentence rationale; otherwise falls back to manual picker with judge's choice preselected
- **`vote`** — multiple judge models evaluate independently, majority wins

Per-branch isolation is configurable (filesystem, memory, todos, message queue, team bus) with safe defaults — each branch gets a virtual filesystem overlay so writes don't collide; only the winner's writes commit to the parent backend on merge.

## Building blocks already in repo or in flight

| Component | Role in forking |
|---|---|
| `CheckpointMiddleware` | Fork point = implicit checkpoint, anchors rollback |
| `AgentTeam` + `TeamMessageBus` | Coordinator pattern + optional inter-branch comms |
| `MessageQueue` (#100) | Per-branch steering via `!{branch_id}` prefix |
| `clone_for_subagent` | Extends → `clone_for_branch` with per-resource isolation flags |
| `for_run()` on capabilities | Per-branch capability state isolation comes for free |
| Esc + `_agent_task` (#96) | Per-branch cancel mechanism reused |

Most of the work is glue, not new infrastructure.

## Proposed file structure

New files added across the six stages — most live under a new `pydantic_deep/toolsets/forking/` package and `apps/cli/widgets/` / `apps/cli/modals/`. Existing files (`pydantic_deep/agent.py`, `pydantic_deep/types.py`, `apps/cli/commands.py`) get small additions, not rewrites.

```
pydantic_deep/
├── capabilities/
│   └── forking.py                       # LiveForkCapability (Stage 1)
├── toolsets/
│   └── forking/
│       ├── __init__.py                  # agent-facing tools: fork_run,
│       │                                #   inspect_branches, merge_or_select,
│       │                                #   terminate_branch, diff_branches, fork_cost
│       ├── coordinator.py               # ForkCoordinator, BranchRuntime         (Stage 1)
│       ├── isolation.py                 # clone_for_branch, BranchOverlay        (Stage 1)
│       ├── store.py                     # ForkStateStore + InMemory impl         (Stage 1)
│       ├── diff.py                      # BranchDiffReport builder + unified diff (Stage 2)
│       ├── materializer.py              # disk flush of branches, manifest       (Stage 5)
│       ├── editor.py                    # detect + invoke pycharm / code diff    (Stage 5)
│       └── judge.py                     # JudgeAgent, confidence scoring, modes  (Stage 6)
├── types.py                             # + BranchSpec, BranchStatus, ForkHandle,
│                                        #   BranchIsolation, MergeStrategy,
│                                        #   PathDiff, BranchChange, DiffSummary,
│                                        #   BranchDiffReport, JudgeVerdict,
│                                        #   ConfidenceSignals, BranchOutcome,
│                                        #   ForkCostSummary
├── agent.py                             # + forking kwarg in create_deep_agent
└── __init__.py                          # re-exports

apps/cli/
├── forking.py                           # /fork, /merge handlers, coordinator bridge  (Stage 3)
├── commands.py                          # + /fork, /merge, /fork open-diff
├── screens/
│   └── chat.py                          # + !{branch_id} input routing
├── widgets/
│   ├── branch_panel.py                  # per-branch streaming view                   (Stage 3)
│   ├── fork_tabs.py                     # tab strip + status badges                   (Stage 3)
│   ├── fork_overview.py                 # all-branches summary                        (Stage 3)
│   ├── merge_acceptance.py              # 2-sentence accept widget                    (Stage 6)
│   └── side_panel.py                    # + ForkBadgeWidget
└── modals/
    ├── fork_picker.py                   # /fork input modal                           (Stage 3)
    └── merge_picker.py                  # /merge modal + judge preselect              (Stages 3 & 6)

tests/
├── test_forking_core.py                 # Stage 1
├── test_forking_diff.py                 # Stage 2
├── test_tui_forking.py                  # Stage 3
├── test_forking_budget.py               # Stage 4
├── test_forking_materializer.py         # Stage 5
├── test_forking_editor.py               # Stage 5
├── test_forking_judge.py                # Stage 6
└── test_tui_forking_acceptance.py       # Stage 6

docs/
└── capabilities/
    └── live-fork.md                     # usage, isolation modes, acceptance modes
```

## Implementation plan

Delivered in six independently mergeable stages (~600–1100 LOC each). Each will be tracked as a sub-issue and linked back here.

- [ ] #102  : coordinator, `BranchOverlay`, `clone_for_branch`, manual merge, 2 branches max. The kernel.
- [ ] #103 : typed `BranchDiffReport` with unified + side-by-side diffs across branches.
- [ ] #104 : `/fork` command, branch tabs, per-branch steering, `/merge` picker.
- [ ] #105 : lift the 2-branch cap, per-branch `budget_usd` enforcement via `CostTracking.for_run()`, aggregate cost view.
- [ ] #106 : live disk materialization of branches, auto-detected `pycharm diff` / `code --diff` invocation, TUI fallback.
- [ ] #107 : `JudgeAgent`, `JudgeVerdict`, confidence scoring, `auto_with_fallback` becomes default. The jump from useful primitive to headline differentiator.

OchnikBartek:
## Summary

Implements **Live Run Forking** — the ability to split an in-flight `agent.run()` into N parallel branches that share history up to the fork point, each diverging with its own steer message. A coordinator resolves the fork via one of four acceptance modes (`manual`, `auto`, `auto_with_fallback` default, `vote`). The winning branch's history is adopted as the parent run's continuation.

Delivered across six stages in a single branch. Closes epic #101.

---

## Added

**Core library (`pydantic_deep/`)**
- `LiveForkCapability` — opt-in capability (`forking=True` on `create_deep_agent`); allocates a `ForkCoordinator` per parent run via `for_run()`
- `ForkCoordinator` — spawns branch tasks, serialises mutations via `asyncio.Lock`, resolves merges, handles partial-history capture on cancellation
- `BranchOverlay` — copy-on-write backend wrapper; reads fall through to parent, writes land in an isolated overlay; `changes()` returns the temporal `list[FileChange]` data spine
- `clone_for_branch` — clones `DeepAgentDeps` per `BranchIsolation` policy (backend, todos, message queue, team bus)
- `InMemoryForkStateStore` / `ForkStateStore` protocol
- `BranchDiffReport` builder (`diff.py`) — typed unified diff across branches with agreement classification (`unanimous_change`, `split`, `unique`)
- `ForkMaterializer` (`materializer.py`) — real-time disk mirror of branch writes under `.pydantic-deep/forks/{fork_id}/`
- `EditorDetector` (`editor.py`) — auto-detects PyCharm / VS Code / custom / TUI fallback for `/fork diff`
- `JudgeAgent` (`judge.py`) — autonomous merge judge with structured `JudgeVerdict`; minimal context (goal + diff report + branch outcomes, no full history)
- `compute_confidence` — `quality_spread × 0.4 + test_pass_ratio × 0.4 + internal_consistency × 0.2`, capped at 0.65 when no test signal
- Test-runner hook — `LiveForkCapability(test_command=..., test_timeout_s=...)` runs a shell command against each branch's materialised snapshot via `asyncio.create_subprocess_exec` (cross-platform, no shell dependency); exit code → `test_pass_ratio`; snapshot I/O runs in thread executor to avoid blocking the asyncio event loop; `UV_NO_SYNC=1` prevents dependency re-resolution in branch snapshot dirs
- Agent-facing tools: `fork_run`, `inspect_branches`, `merge_or_select`, `terminate_branch`, `diff_branches`, `fork_cost`
- New types: `BranchSpec`, `BranchIsolation`, `BranchStatus`, `ForkHandle`, `MergeStrategy`, `BranchDiffReport`, `PathDiff`, `BranchChange`, `DiffSummary`, `JudgeVerdict`, `ConfidenceSignals`, `BranchOutcome`, `ResolveOutcome`, `BranchCost`, `ForkCostSummary`, `FlushError`, `FlushReport`

**CLI (`apps/cli/`)**
- `CLIForkSession` (`forking.py`) — bridges coordinator to CLI; `run_on_branch`, `merge`, `inspect`, `steer_branch`; `reconcile_active_fork` auto-adopts agent-initiated coordinators after parent turn ends (stash + adopt mechanics)
- `/fork` command + `ForkPickerModal` — collect N `(label, steer, model, budget)` tuples, spawn branches
- `/merge` command + `MergePickerModal` — manual pick or judge-preselected with reasoning subtitle
- `ForkConfigModal` (`modals/fork_config.py`) — `/fork-config` settings modal; resizable branch-count input remounts per-branch model/budget rows; persists all values to `.pydantic-deep/config.toml` on Save
- `DiffPickerModal` (`modals/diff_picker.py`) — `/fork diff` path + branch selector; lists every touched path from `BranchDiffReport`, lets user toggle which branches to include before launching the external diff tool
- `BranchApprovalModal` (`modals/branch_approval.py`) — tool-approval gate shown when a running branch has a pending approval request; poll timer surfaces it, answer forwarded to the branch's queue to unblock it
- `BranchPanelWidget` — per-branch streaming view with incremental `replay_messages_append` (watermark-based delta rendering)
- `ForkTabsWidget` — tab strip with live status badges (● / ✓ / ✗ / ⊘)
- `ForkOverviewWidget` — all-branches summary
- `ForkBadgeWidget` — side-panel aggregate cost + status chip
- `JudgeLoadingScreen` — modal spinner overlay during judge evaluation; appears immediately on `/merge` before judge LLM call; Esc aborts
- `MergeAcceptanceWidget` — `auto_with_fallback` accept/diff/override widget
- `>>label msg` per-branch steering (extends Message Queue #100 routing)
- Interactive branch chat (E2): plain text on a focused `done` branch tab routes to `run_on_branch`; per-branch `extra_message_history` accumulates independent follow-up turns
- Live streaming via `branch_runner` / `agent.iter()` — token-by-token output in branch panels
- `/fork diff`, `/fork diff <path>` — diff explorer with external tool invocation
- `/fork-branches N`, `/fork-budget`, `/fork-model` configuration commands (persisted to `.pydantic-deep/config.toml`)

**Tests** — 18 new test files, +21 384 lines total, 100% line + branch coverage on all new modules

**Docs** — `docs/capabilities/live-fork.md` full reference: isolation flags, acceptance modes, confidence scoring, test-runner hook, diff explorer, IDE integration, CLI commands, interactive branch chat, stash + adopt mechanics

---

## Changed

- `pydantic_deep/agent.py` — added `forking: bool | LiveForkCapability = False` to `create_deep_agent()`
- `pydantic_deep/types.py` — extended with all forking types listed above
- `pydantic_deep/__init__.py` — re-exports for all public forking symbols
- `pydantic_deep/prompts.py` — added `## Forking` section to `BASE_PROMPT` with poll-via-`inspect_branches` protocol and `merge_or_select` resolution contract so agents don't call `wait_tasks` on branch ids
- `apps/cli/app.py` — added fork reactives (`active_fork`, `fork_branch_count`, `fork_branch_models`, `fork_branch_budgets`, `fork_merge_strategy`, `fork_judge_model`, `fork_confidence_threshold`) and `_seed_fork_settings_from_config` called on mount to restore persisted settings
- `apps/cli/config.py` — added fork config fields (`fork_branch_count`, `fork_aggregate_budget_usd`, `fork_branch_models`, `fork_branch_budgets`) with TOML serialisation; extended `_INT_FIELDS` / `_FLOAT_FIELDS` type coercion
- `apps/cli/commands.py` — wired `/fork`, `/merge`, `/fork diff`, `/fork-branches`, `/fork-budget`, `/fork-model`, `/fork-config`
- `apps/cli/screens/chat.py` — added `_poll_fork_state`, `enter_fork_view`, `_stream_branch_via_iter`, interactive branch chat routing in `on_user_submitted`

---

## Testing

Manual CLI testing: `uv run pydantic-deep -w fork_tests` — 10 test cases in `fork_tests/PROMPTS.md` covering manual merge, judge modes, `>>` steering, diff explorer, terminate/abort, interactive branch chat, agent-initiated fork, stash + adopt, and budget enforcement.

---

## Related Issues

Closes #101  
Stages: #102 · #103 · #104 · #105 · #106 · #107

---

## Notes for Reviewers

**Isolation defaults** — `BranchIsolation` defaults are `backend="copy"`, `message_queue="isolated"`, `team_bus="shared"`. The asymmetry is intentional: per-branch `>>label` steering requires queue isolation, but branches communicating over the shared team bus is desirable for cooperative tasks.

**`JudgeAgent` is not a subagent** — fixed internal helper inside `ForkCoordinator.resolve()`, not registered with `SubAgentToolset`.

**`combine` mode is explicitly absent** from `MergeStrategy.kind` — intentional, meaningful only for non-code outputs.

**Known issues being tracked as follow-up:**
1. Abort via Esc on overview does not remove branch panels — ghost fork panel can re-appear on next parent turn
2. Interactive branch chat: message renders only after run completes — streaming path loses panel reference on subsequent `run_on_branch` calls
3. `done → running` status flip not reflected in `ForkTabsWidget` / `ForkOverviewWidget` until next poll tick


bll-kacper-wlodarczyk (commented):
## Automated review — Live Run Forking (PR #123)

High-effort recall pass (7 finder angles → verification). **10 confirmed/plausible findings** posted inline below, ranked by severity: 3 security/correctness regressions, 4 concurrency/resource bugs, 3 correctness crashes.

**Refuted candidates (not posted, for the record):** `judge.py` `result.usage` — actually a callable `RunUsage` subclass via `deprecated_callable_property`, so it works; `diff.py` `agreement_score` — a `split` path always has ≥2 touchers so it is always ⊆ `union_touched`, ratio can't exceed 1; `app.py` Esc handler — falls through to `focus_input`, not `self.exit()`; `forking=True, cost_tracking=False` budgets — still enforced via the `_build_branch_cost_tracking` model-name fallback + independent aggregate watcher.

**Additional lower-priority items worth tracking (not blocking):**
- `coordinator.py:1387` `capture_partial_history` does `list(messages)` (full copy) on *every* `before_model_request` → O(N²) copying over a branch run; consider append/tail-delta.
- `isolation.py:540` `_branch_snapshot` rebuilds a full recursive symlink mirror of the whole project tree + walks it twice on *every* `execute()` call (no caching) → O(files) per shell command, ×N branches.
- `coordinator.py:231` per-branch `_BudgetWatcher` self-cancels its own running task from inside the cost callback (fired in `after_run`); the cancel may be a no-op so an over-budget branch can finish `done` instead of `budget_exhausted`.

---
_Generated with Claude Code._

pydantic_deep/agent.py
```diff
@@ -582,7 +600,11 @@ def _set_toolset_retries(toolset: AbstractToolset[DeepAgentDeps], max_retries: i
         require_write_approval = interrupt_on.get("write_file", False) or interrupt_on.get(
             "edit_file", False
         )
-        require_execute_approval = interrupt_on.get("execute", True)
+        # Default False: only require approval when explicitly configured.
+        # Default True caused a broken state — execute deferred but DeferredToolRequests
+        # never added to output types (has_interrupt_tools checked any(interrupt_on.values())
+        # which is False for an empty dict). Use explicit {"execute": True} to require approval.
+        require_execute_approval = interrupt_on.get("execute", False)
```
DEENUU1: 🔴 **Security regression — execute approval gate silently dropped.** The default flipped from `interrupt_on.get("execute", True)` to `... False`. The PR rationale (old `execute: True` default was broken when `interrupt_on` was empty) only holds for the *empty-dict* case — there `has_interrupt_tools` is False so `DeferredToolRequests` was never wired and execute deferral had no channel anyway.

But for a **non-empty** `interrupt_on` that omits `execute` — e.g. `create_deep_agent(interrupt_on={"edit_file": True})` (exactly what `tests/test_agent_extended.py` uses) — the old default gated shell `execute` behind human approval *and* `DeferredToolRequests` was added. Now `require_execute_approval=False`, so edit/write still pause but **arbitrary shell commands run unattended**. A caller who enabled edit interrupts and relied on execute also being gated loses that gate with no warning.

Suggest: keep `execute` defaulting to gated when any interrupt is enabled, or document the behavior change loudly.

pydantic_deep/toolsets/forking/materializer.py
```diff
@@ -0,0 +1,158 @@
...
+        self.manifest_path().write_text(json.dumps(payload, indent=2), encoding="utf-8")
+
+
+def _safe_relative(path: str) -> Path:
+    """Return ``path`` as a :class:`Path` rooted relative to materializer dirs.
+
+    Strips a leading ``/`` so absolute-style paths (``/src/foo.py``) land
+    under the materializer root instead of escaping to the filesystem root.
+    """
+    return Path(path.lstrip("/"))
```
DEENUU1: 🔴 **Path traversal — `_safe_relative` doesn't neutralize `..`.** `Path(path.lstrip("/"))` strips a leading slash but leaves `..` segments intact, so a write path like `../../etc/cron.d/x` resolves to `self.root / 'branches' / label / Path('../../...')` **outside** the fork directory. A branch agent that writes such a path (`BranchOverlay.write` → `_mirror_to_disk` → `flush_change` → `_safe_relative`) lands `write_bytes`/`chmod` on an arbitrary filesystem location, and `cleanup()`'s `rmtree(self.root)` won't remove it.

Suggest: drop `..` components / resolve and confine under `root`, e.g. reject or strip any segment `== '..'` and assert the resolved path is relative to `self.root`.

pydantic_deep/toolsets/forking/isolation.py
```diff
@@ -0,0 +1,937 @@
...
+    effect here. ``team_bus`` is a no-op when the teams capability is not
+    enabled on the parent run; when enabled it propagates the parent bus
+    reference by default.
+    """
+
+    new_backend: BackendProtocol = (
+        BranchOverlay(deps.backend) if isolation.backend == "copy" else deps.backend
+    )
+
+    new_todos = [] if isolation.todos == "copy" else deps.todos
```
DEENUU1: 🔴 **`todos="copy"` is doubly wrong.** `new_todos = [] if isolation.todos == "copy" else deps.todos`:

1. The `"copy"` branch (the documented default — `BranchIsolation.todos` docstring says "copy") returns an **empty list**, not `list(deps.todos)`. If the user forks after several planning turns, every branch starts with zero todos and loses the parent's in-progress plan.
2. The `else` branch shares the **same list object by reference**, so concurrent branch tasks' `write_todos` mutate the parent's and siblings' lists — a lost-update race (read → await model call → sibling mutates → stale write-back).

Suggest: `new_todos = list(deps.todos) if isolation.todos == "copy" else deps.todos`, and confirm the `share` path's reference-sharing is intended under concurrency.

pydantic_deep/toolsets/forking/__init__.py
```diff
@@ -0,0 +1,471 @@
...
+                action = f"pick:{outcome.verdict.winner_branch_id}"
+            else:
+                return (
+                    "merge_or_select auto: strategy is 'manual' — "
+                    "call merge_or_select(action='pick:<branch_id>') explicitly."
+                )
+
+        try:
+            result = await coordinator.merge_or_select(action)
+        except ValueError as e:
```
DEENUU1: 🔴 **`merge_or_select` tool catches only `ValueError`, but the coordinator raises `RuntimeError`.** `coordinator.merge_or_select` raises `RuntimeError` when the picked winner was cancelled/budget-exhausted before merge (`coordinator.py:861`) or when called before `fork()` (`coordinator.py:843`), and `await winner.task` (`coordinator.py:848`) re-raises a failed branch's own exception. None of these are `ValueError`, so they escape this `except ValueError` and **abort `agent.run()`** instead of returning a graceful error string.

Repro: `auto_with_fallback`, judge recommends branch B, B hits its `budget_usd` and is terminated between verdict and commit → the auto path falls through to `pick:<B>` → `RuntimeError` propagates out of the tool and crashes the run; `reconcile_active_fork` then keeps re-adopting a coordinator that can never resolve.

Suggest: `except (ValueError, RuntimeError)` (and consider catching the branch exception from `await winner.task`).

pydantic_deep/toolsets/forking/coordinator.py
```diff
@@ -0,0 +1,1492 @@
...
+            raise ValueError("pick action missing branch id.")
+        if target_id not in self.branches:
+            raise ValueError(f"Unknown branch id: {target_id!r}")
+        if self._handle is None:  # pragma: no cover - defensive
+            raise RuntimeError("merge_or_select called before fork()")
+
+        async with self._lock:
+            winner = self.branches[target_id]
+            try:
+                result = await winner.task
```
DEENUU1: 🔴 **Deadlock: `await winner.task` while holding `self._lock`.** This `await` (line 848) is inside `async with self._lock` (line 845). If the winning branch is parked on a pending tool-approval (`await request.response.get()`) the user never answers — or is just a long/stuck run — the lock is held for that entire (unbounded, human-in-the-loop) duration. Every other lock user then hangs: `fork()` (line 476), `run_on_branch()` (line 692), the abort path, etc.

Suggest: snapshot what you need under the lock, then `await winner.task` *outside* the `async with`, or use a per-branch lock so a blocked winner can't freeze the whole coordinator.

pydantic_deep/toolsets/forking/coordinator.py
```diff
@@ -0,0 +1,1492 @@
...
+    async def aclose(self) -> None:
+        """Cancel every outstanding branch task — used on parent cancellation.
+
+        Also runs ``materializer.cleanup()`` so the on-disk fork directory
+        is removed on abort (unless ``keep_artifacts`` is set), mirroring
+        the merge-resolution cleanup. Safe to call multiple times.
+        """
+        for rt in self.branches.values():
+            if not rt.task.done():
+                rt.task.cancel()
```
DEENUU1, line 1452: 🟠 **`aclose()` rmtree's the fork dir without awaiting cancelled tasks.** It calls `rt.task.cancel()` (line 1398) then immediately `materializer.cleanup()` (`shutil.rmtree`, line 1400) — but never awaits the tasks, so a branch still unwinding mid-write (`BranchOverlay.write` → `materializer.flush_change` → `mkdir`/`write_bytes`) can run **after** the rmtree and recreate part of `.pydantic-deep/forks/{fork_id}`, leaking an orphaned dir (these accumulate across aborted parent runs with `keep_artifacts=False`).

Contrast `merge_or_select` (lines 890–892) and `abort_fork` (lines 946–949), which `await asyncio.wait_for(rt.task, ...)` before cleanup. `aclose` should do the same.

DEENUU1 (commented):
## Automated review — Live Run Forking (PR #123)

High-effort recall pass (7 finder angles → verification). **10 confirmed/plausible findings** posted inline below, ranked by severity: 3 security/correctness regressions, 4 concurrency/resource bugs, 3 correctness crashes.

**Refuted candidates (not posted, for the record):** `judge.py` `result.usage` — actually a callable `RunUsage` subclass via `deprecated_callable_property`, so it works; `diff.py` `agreement_score` — a `split` path always has ≥2 touchers so it is always ⊆ `union_touched`, ratio can't exceed 1; `app.py` Esc handler — falls through to `focus_input`, not `self.exit()`; `forking=True, cost_tracking=False` budgets — still enforced via the `_build_branch_cost_tracking` model-name fallback + independent aggregate watcher.

**Additional lower-priority items worth tracking (not blocking):**
- `coordinator.py:1387` `capture_partial_history` does `list(messages)` (full copy) on *every* `before_model_request` → O(N²) copying over a branch run; consider append/tail-delta.
- `isolation.py:540` `_branch_snapshot` rebuilds a full recursive symlink mirror of the whole project tree + walks it twice on *every* `execute()` call (no caching) → O(files) per shell command, ×N branches.
- `coordinator.py:231` per-branch `_BudgetWatcher` self-cancels its own running task from inside the cost callback (fired in `after_run`); the cancel may be a no-op so an over-budget branch can finish `done` instead of `budget_exhausted`.

---
_Generated with Claude Code._

pydantic_deep/toolsets/forking/coordinator.py
```diff
@@ -0,0 +1,1492 @@
...
+                )
+            except (OSError, FileNotFoundError) as exc:
+                logger.warning(
+                    "branch %s test command failed to spawn: %s",
+                    rt.status.id,
+                    exc,
+                )
+                return None
+            try:
+                returncode = await asyncio.wait_for(proc.wait(), timeout=self.test_timeout_s)
```
DEENUU1, line 1094: 🟠 **Test subprocess leaks on cancellation.** `proc.terminate()`/`kill()` only runs inside the `except asyncio.TimeoutError` branch (lines 1038–1047). If the task is cancelled (parent abort during the auto-resolve test phase) while awaiting `proc.wait()` at line 1036, `CancelledError` propagates with the child process **never terminated** → zombie/leaked process. Additionally the `finally` awaits `run_in_executor(snap_ctx.__exit__)`, which is itself cancellable during unwind → the snapshot tempdir may not be cleaned.

Suggest: wrap the subprocess in `try/finally` that terminates/kills `proc` on any exit (including `CancelledError`), and shield the tempdir cleanup.

pydantic_deep/toolsets/forking/coordinator.py
```diff
@@ -0,0 +1,1492 @@
...
+            flush_errors: list[Any] = []
+            deleted_paths: list[str] = []
+            winner_overlay = winner.overlay
+            if winner_overlay is not None:
+                snapshot = (
+                    self.materializer.pre_flush_snapshot()
+                    if self.materializer is not None
+                    else None
+                )
+                report = winner_overlay.flush_to(self.parent_deps.backend, snapshot)
```
DEENUU1, line 949: 🟠 **Copy-on-write read isolation violated: winner flushed before losers cancelled.** `winner_overlay.flush_to(parent)` (line 877) writes the winner's changes to the **shared parent backend** before the loop that cancels losers (lines 889+). `cancel()` only schedules cancellation (effective at the loser's next `await`), and `BranchOverlay` reads fall through to the parent — so a not-yet-cancelled loser that reads a path it didn't itself touch now sees the **winner's just-flushed bytes** instead of the pre-fork parent state. Their tool results / `partial_history` (used by a re-resolve or the judge) can be computed from cross-branch state.

Suggest: cancel + await all losers to quiescence *before* flushing the winner.

apps/cli/config.py
```diff
@@ -262,9 +342,26 @@ def _coerce_value(key: str, value: str) -> Any:
     if key in _INT_FIELDS:
         return int(value)
     if key in _FLOAT_FIELDS:
+        if value.lower() in ("none", "null", ""):
+            return None
```
DEENUU1: 🟠 **`None` coercion crashes the non-optional `fork_confidence_threshold`.** This new branch maps empty/`none`/`null` → `None` for *all* `_FLOAT_FIELDS`, but `fork_confidence_threshold` is declared `float = 0.80` (non-optional, see line 180) and is in `_FLOAT_FIELDS` (line 88). `/set fork_confidence_threshold none` (or empty) stores `None`, then `validate_config` does `config.fork_confidence_threshold < 0.0` (line 275) → `TypeError: '<' not supported between 'NoneType' and 'float'`.

Suggest: gate the `None` coercion to genuinely-optional float fields (e.g. an explicit `_OPTIONAL_FLOAT_FIELDS` set), not the whole `_FLOAT_FIELDS`.

pydantic_deep/toolsets/liteparse.py
```diff
@@ -11,27 +11,31 @@
...
 except ImportError:  # pragma: no cover
+    pass
 
-    class _LiteParse:  # type: ignore[no-redef]
-        def __init__(self, **kwargs: Any) -> None: ...  # pragma: no cover
-
-    class _CLINotFoundError(Exception):  # type: ignore[no-redef]
-        pass  # pragma: no cover
+with contextlib.suppress(ImportError, AttributeError):
+    _CLINotFoundError = importlib.import_module("liteparse.types").CLINotFoundError
```
DEENUU1: 🟡 **Synthetic `_CLINotFoundError` breaks the friendly-error path under version skew.** `_HAS_LITEPARSE` is now set `True` from the `LiteParse` import alone, while `CLINotFoundError` is imported separately under `contextlib.suppress(ImportError, AttributeError)` (lines 40–41) with a synthetic fallback `type("_CLINotFoundError", (Exception,), {})` (line 24). If liteparse is installed but `liteparse.types.CLINotFoundError` is missing/renamed (version skew), `_CLINotFoundError` stays the synthetic dummy — and the real `except _CLINotFoundError` handlers (lines 141, 195) and the exported `LiteparseCliNotFoundError` alias **no longer catch the real exception**, so the `_NOT_INSTALLED_MSG` path is bypassed and a raw error propagates.

Previously this import lived in the same `try` as `LiteParse`, so a missing `CLINotFoundError` cleanly set `_HAS_LITEPARSE = False`. Suggest tying `_HAS_LITEPARSE` to both imports succeeding.

DEENUU1 (commented):
## Automated review — batch 2 (deeper sweep)

Follow-up pass focused on the surface the first review under-covered: `apps/cli/` (commands, chat, forking, widgets, modals) and a second deep pass over `isolation.py` / `coordinator.py` / `diff.py` / `judge.py`. **18 more confirmed/plausible findings below**, including two critical isolation/data-loss bugs (symlink writes escaping COW; conflict-clobber on the parent). Each was checked by an independent verifier.

**Refuted this round (not posted):** `commands.py` acceptance-widget `verdict` deref — provably non-None on that path (only reached when `auto_eligible`, which only `auto_with_fallback` produces, and there `verdict` is the judge result); `merge_picker.py` wrong-label — the `status.label` fallback restores the correct label.

Test files (+12k lines) were intentionally skipped — not shipped. Happy to review them for assertion gaps if useful.

---
_Generated with Claude Code._

pydantic_deep/toolsets/forking/isolation.py
```diff
@@ -0,0 +1,937 @@
...
+            if entry.name in _SNAP_SKIP_DIRS:
+                continue
+            target = dst / entry.name
+            if entry.is_dir(follow_symlinks=False):
+                target.mkdir(exist_ok=True)
+                _symlink_tree(Path(entry.path), target)
+            else:
+                # Symlink to absolute path so the link stays valid even
+                # when cwd changes inside the subprocess.
+                target.symlink_to(Path(entry.path).resolve())
```
DEENUU1: 🔴🔴 **CRITICAL — symlink snapshot lets branch writes escape copy-on-write onto the real parent.** `_symlink_tree` mirrors untouched parent files as absolute symlinks (`target.symlink_to(Path(entry.path).resolve())`). The branch's `execute` then runs `sh -c <cmd>` against this snapshot (line ~545) with **no write interception** — only a post-hoc mtime diff. Any in-place write follows the symlink straight to the parent file: `echo x >> src/app.py`, `sed -i`, a truncating rewrite, an editor that doesn't atomic-rename. So a *losing* branch's side effects (and the `test_command`'s) leak into the real parent and become visible to other branches **before any merge/winner selection** — defeating the core isolation guarantee of the whole feature.

Suggest: copy-on-write at the file level (copy on first write, or run the snapshot in a sandbox/overlayfs/bind-mounted tmp), not bare symlinks; at minimum detach+copy a symlinked file before exposing it to a writable execute.

pydantic_deep/toolsets/forking/isolation.py
```diff
@@ -0,0 +1,937 @@
...
+                        applied_set.add(change.path)
+                        applied_paths.append(change.path)
+                continue
+            if change.op == "rmdir":
+                if self._flush_rmdir(parent, change, errors, deleted_paths, deleted_set):
+                    applied_changes += 1
+                continue
+            content = self._overlay.read_bytes(change.path)
+            try:
+                write_result = parent.write(change.path, content)
```
DEENUU1, line 749: 🔴 **Detected merge conflict is reported but the parent is clobbered anyway (silent data loss).** `flush_to` computes `conflicts = self._detect_conflicts(parent, pre_flush_snapshot)` (line 676) — paths where the parent changed since fork — but the flush loop still calls `parent.write(change.path, content)` for every change (line 700) with no skip on conflict and no backup. If a third actor (concurrent agent, the user saving, another tool) modified parent file P between fork and merge, the winning branch's flush **overwrites P's newer content**; P is merely listed in `conflicts` after the fact, with the conflicting version already destroyed.

Suggest: skip (or three-way merge / back up) conflicting paths instead of last-write-wins, and surface the conflict to the user *before* writing.

pydantic_deep/toolsets/forking/coordinator.py
```diff
@@ -0,0 +1,1492 @@
...
+                    f"Branch {branch_id!r} is still running — cannot start a new turn."
+                )
+            if runtime.status.state != "done":
+                raise RuntimeError(
+                    f"Branch {branch_id!r} is in state {runtime.status.state!r} — "
+                    "only 'done' branches accept new turns."
+                )
+            prev_result = runtime.task.result()
+            base_history = list(prev_result.all_messages())
+            effective_history = base_history + list(runtime.extra_message_history)
```
DEENUU1: 🔴 **`run_on_branch` double-counts history on every continued turn.** `base_history = prev_result.all_messages()` already contains the previous turn's tail, yet `effective_history = base_history + runtime.extra_message_history` adds it again — because the done-callback (line 738) extends `extra_message_history` with `tail = new_messages[len(effective_history):]`.

Trace: turn 1 ends → `extra=[tail1]`. Turn 2: `base = all_messages() = seed+tail0+tail1`, `effective = base + extra = seed+tail0+tail1+**tail1**`. The duplicate is fed back to the model, and `merge_or_select` (line 850) does `result.all_messages() + winner.extra_message_history`, duplicating once more in the adopted history. Conversation grows with duplicated request/response pairs after any second branch turn.

Suggest: `extra_message_history` is redundant with `all_messages()` — use one or the other. Either seed from `base_history` only, or track turns purely via `extra` against a fixed seed (not `all_messages()`).

pydantic_deep/toolsets/forking/isolation.py
```diff
@@ -0,0 +1,937 @@
...
+        if parent_root is None:
+            return self._parent.execute(command, timeout)  # pyright: ignore[reportAttributeAccessIssue]
+
+        parent_root = Path(parent_root)
+        timeout_s = timeout if timeout is not None else _EXEC_DEFAULT_TIMEOUT_S
+
+        with _branch_snapshot(parent_root, self._overlay, self._changes, self._deleted) as snap:
+            snap_path = Path(snap)
+            pre = _snapshot_state(snap_path)
+            cmd = command.replace(str(parent_root), snap)
```
DEENUU1: 🟠 **Snapshot path rewrite is a naive substring replace.** `cmd = command.replace(str(parent_root), snap)` rewrites *any* substring match of the parent root, not path-boundary matches. With `parent_root=/home/u/proj`, a command touching a sibling like `/home/u/proj_backup/x` becomes `<snap>_backup/x` — a path that doesn't exist in the snapshot — silently corrupting the branch's command (and thus its diff/test signals). Conversely the root appearing inside an unrelated string literal gets mangled.

Suggest: set the subprocess `cwd` to the snapshot and pass relative paths, or rewrite only on path boundaries — don't string-substitute into arbitrary shell text.

pydantic_deep/toolsets/forking/isolation.py
```diff
@@ -0,0 +1,937 @@
...
+
+        if not in_pre and in_post:
+            _capture_overlay_write(overlay, abs_path, snap_file)
+            continue
+
+        pre_sym, pre_mtime = pre[rel]
+        post_sym, post_mtime = post[rel]
+
+        symlink_replaced_by_file = pre_sym and not post_sym
+        real_file_changed = not post_sym and pre_mtime != post_mtime
```
DEENUU1: 🟠 **Mutation detection is mtime-only — writes within mtime resolution are lost.** `real_file_changed = not post_sym and pre_mtime != post_mtime` (and `_snapshot_state` stores only `(is_symlink, mtime)`). A command that rewrites a file within the filesystem's coarse mtime tick, or a tool that preserves mtime (`cp -p`, `touch -r`, some formatters), leaves `pre_mtime == post_mtime` → `real_file_changed=False` → the overlay never records the change. The branch's actual on-disk edit is then absent from the diff and never flushed even if the branch wins.

Suggest: compare size + content hash (or at least size) as a fallback, not mtime alone.

pydantic_deep/toolsets/forking/diff.py
```diff
@@ -0,0 +1,353 @@
...
+    touchers = [bc for bc in branches.values() if bc.operation != "untouched"]
+    if not touchers:
+        return "unanimous_no_change"
+    if len(touchers) == 1:
+        return "unique"
+    first = touchers[0]
+    for other in touchers[1:]:
+        if (
+            other.operation != first.operation
+            or other.new_content != first.new_content
```
DEENUU1: 🟠 **Divergent binary writes are misclassified as agreement.** `_classify_agreement` keys on `(operation, new_content, is_binary)`, but binary `BranchChange`s carry `new_content=None` (content isn't captured for binary, lines 185/205/228). Two branches writing *different* binary bytes to the same path (e.g. `asset.bin`) both produce `(modified, None, True)` → classified `unanimous_change` instead of `split`. The real conflict is invisible in the diff report, inflates `agreement_score`, and the judge/confidence pipeline treats it as full agreement.

Suggest: compare a content hash for binary paths (the hash is already computed for the unified-diff placeholder elsewhere).

pydantic_deep/toolsets/forking/coordinator.py
```diff
@@ -0,0 +1,1492 @@
...
+                auto_eligible=False,
+                verdict=None,
+                signals=None,
+                effective_confidence=0.0,
+                merge_result=None,
+                judge_usage=None,
+            )
+
+        # Skip re-invoking the judge when the user re-opens /merge with the same strategy.
+        if (
```
DEENUU1: 🟠 **`resolve()` caches the judge outcome keyed only on `strategy.kind`.** The cache check (~line 1210) compares only `_cached_outcome_strategy_kind`, ignoring `MergeStrategy.confidence_threshold`, `judge_model`, and `judge_models`. A second `/merge` with the same kind but a different threshold or judge panel returns the **stale** `ResolveOutcome` — `auto_eligible`/`effective_confidence` reflect the old threshold and the new judge models are never consulted (the cache is only invalidated on a new turn/fork, not on re-resolve).

Suggest: include threshold + judge model(s) in the cache key, or invalidate when the strategy differs in any field.

pydantic_deep/toolsets/forking/judge.py
```diff
@@ -0,0 +1,317 @@
...
+    ``rejected_with_reasons`` taking the highest-confidence entry per branch).
+    """
+    if not verdicts:
+        raise ValueError("_majority_pick requires at least one verdict.")
+    counts = Counter(v.winner_branch_id for v in verdicts)
+    top_count = max(counts.values())
+    tied_winners = [bid for bid, c in counts.items() if c == top_count]
+    candidates = [v for v in verdicts if v.winner_branch_id in tied_winners]
+    # Highest-confidence verdict among the tied set is the tie-breaker.
+    best = max(candidates, key=lambda v: v.confidence)
```
DEENUU1, line 205: 🟡 **Judge `confidence` is unbounded and unvalidated.** `JudgeVerdict.confidence: float` (types.py:454) has no `Field(ge=0, le=1)`, and `evaluate` returns `result.output` without clamping. A model emitting `confidence=1.5` skews `_majority_pick`'s tie-break `max(..., key=v.confidence)` (line 205) and inflates `mean_confidence` (line 208) and `compute_confidence`'s `heuristic * judge_confidence` product (only the *final* product is clamped, so the surfaced confidence and tie-break ordering are still wrong).

Suggest: constrain the field (`ge=0, le=1`) or clamp on ingestion.

apps/cli/widgets/branch_panel.py
```diff
@@ -0,0 +1,302 @@
...
+            self.query_one(".branch-header", Static).update(self._render_header())
+
+    def watch_awaiting_approval(self, _old: bool, _new: bool) -> None:
+        with contextlib.suppress(Exception):  # widget not yet mounted
+            self.query_one(".branch-header", Static).update(self._render_header())
+
+    def mark_status(self, state: BranchState, reason: str | None = None) -> None:
+        """Update the branch's status badge and optional reason text."""
+        if state == "running":
+            self._last_replayed_len = 0
```
DEENUU1, line 166: 🟠 **`mark_status('running')` resets the replay watermark without clearing the rendered list → duplicated transcript.** It sets `_last_replayed_len = 0` and clears `_rendered_call_ids` (line 161) but never calls `msg_list.clear_messages()`. `replay_messages_append` slices `messages[_last_replayed_len:]`, so a non-running→running transition on a poll tick followed by `replay_messages_append(partial_history)` re-renders every message from index 0 into the still-populated list, duplicating the whole branch transcript. This is reachable via the `done→running` flip the PR itself lists as a known issue, and via `run_on_branch` continued turns.

Suggest: clear the MessageList when resetting the watermark (or don't reset the watermark on status change).

apps/cli/widgets/branch_panel.py
```diff
@@ -0,0 +1,302 @@
...
+                    args = part.args_as_dict()
+                    assistant_msg = msg_list.current_assistant
+                    if assistant_msg is None:
+                        assistant_msg = msg_list.begin_assistant_message()
+                    assistant_msg.add_tool_call(part.tool_name, args, call_id)
+                    self._rendered_call_ids.add(call_id)
+                elif isinstance(part, ToolReturnPart):
+                    content_str = str(part.content)
+                    assistant_msg = msg_list.current_assistant
+                    if assistant_msg is not None:
```
DEENUU1: 🟠 **Tool call left stuck 'pending' when its return straddles a poll tick.** In `replay_messages_append`, a `ToolReturnPart` is only completed `if assistant_msg is not None` (line 287). `current_assistant` is set to `None` by `end_assistant_message()` whenever a slice ends with a `TextPart`. So if tick N's slice ends with a text part and tick N+1 begins with the `ToolReturnPart` for a call rendered in tick N, the guard skips `complete_tool_call` and the tool row spins forever despite having returned.

Suggest: look the call up by `tool_call_id` across rendered messages instead of relying on `current_assistant`.

apps/cli/widgets/branch_panel.py
```diff
@@ -0,0 +1,302 @@
...
+                        assistant_msg.complete_tool_call(
```
DEENUU1: 🟡 **Tool result can attach to the wrong assistant message across ticks.** `complete_tool_call` (line 288) is called on the *current* assistant message, not the one that actually holds the matching `ToolCallPart`. When a call-part (tick N, assistant A) and its return-part (tick N+1, a new assistant B) span a tick boundary, the result is completed against B (or dropped), so it renders under the wrong tool-call row. Same root cause as the stuck-pending bug above — resolve both by keying on `tool_call_id`.

apps/cli/screens/chat.py
```diff
@@ -384,13 +523,101 @@ def _show_welcome(self) -> None:
...
         """Handle user submitting a prompt."""
+        import re
+
         text = event.text
 
         app = self.app
+        active_fork = app.active_fork
+
+        if active_fork is not None:
+            match = re.match(r"^>>(\w+)\s+(.*)", text)
```
DEENUU1: 🟠 **`>>label` steering silently breaks for labels with hyphens or spaces.** The regex `^>>(\w+)\s+(.*)` uses `\w+`, which excludes `-` and space. But `ForkPickerModal` accepts any non-empty distinct label (no charset restriction), so a branch labeled `approach-a` can't be steered: the regex fails to match, and the message falls through to the general handler (never delivered to the branch queue). A label like `fix bug` matches `fix` and routes to a nonexistent branch.

Suggest: broaden the capture (`^>>(\S+)\s+(.*)`) and resolve against actual labels, or restrict labels to `\w+` at creation time for consistency.

apps/cli/forking.py
```diff
@@ -0,0 +1,349 @@
...
+        parent_history=safe_history,
+        isolation=isolation,
+        aggregate_budget_usd=result.aggregate_budget_usd,
+        strategy=merge_strategy,
+    )
+    label_to_id: dict[str, str] = {}
+    for branch_id in handle.branches:
+        runtime = coordinator.branches[branch_id]
+        label_to_id[runtime.spec.label] = branch_id
+    return CLIForkSession(coordinator=coordinator, handle=handle, label_to_id=label_to_id)
```
DEENUU1, line 252: 🟠 **Duplicate/empty branch labels collapse `label_to_id`, making branches unreachable.** `label_to_id[runtime.spec.label] = branch_id` (here and in `adopt_agent_coordinator` ~329) has no uniqueness/non-empty guard. The CLI picker enforces distinct labels, but the agent-facing `fork_run` tool passes labels through verbatim and the coordinator's `fork()` doesn't enforce uniqueness (`BranchSpec` has no validation). Two branches sharing a label → last-write-wins in the map → `_resolve_id(label)` resolves both `>>label` steers to one branch and the other is unreachable / reports the wrong `branch_state`. (`diff_picker.py:318` inverts the same map and collapses too.)

Suggest: enforce unique non-empty labels at `fork()` (auto-suffix dupes), and key the map on branch id with label as display-only.

apps/cli/commands.py
```diff
@@ -602,3 +613,641 @@ async def _handle_key(
...
+        return
+
+    kind = EditorDetector.detect()
+    if kind == "tui":
+        report = session.build_diff()
+        if report is None:  # pragma: no cover - defensive
+            app.notify("Cannot build diff report", severity="error")
+            return
+        statuses = session.inspect()
+        app.push_screen(MergePickerModal(report, statuses, session.label_to_id))
```
DEENUU1: 🟠 **`/fork diff` TUI fallback is a silent no-op that looks like a merge.** When no external difftool is detected, this pushes `MergePickerModal(...)` without `view_only=True` and without a result callback. The modal renders the full "pick a winner" hint and binds Enter/number keys to `action_pick_*` → `dismiss(MergePickerResult(...))`, but since `push_screen` has no callback the result is discarded. The user presses Enter expecting to merge/select the inspected branch and **nothing happens**.

Suggest: pass `view_only=True` for the browse path (Enter just closes), or wire a callback that performs the selection.

apps/cli/modals/fork_config.py
```diff
@@ -0,0 +1,521 @@
...
+                self._show_error(
+                    f"Branch {i + 1} model {slot!r} missing provider prefix "
+                    "(e.g. 'anthropic:claude-opus-4-6')"
+                )
+                return
+            models.append(slot)
+
+        budgets: list[float | None] = []
+        for i in range(count):
+            raw = self._input(f"fork-config-budget-{i}").value.strip()
```
DEENUU1: 🟡 **`/fork-config` Save crashes if a per-branch row isn't mounted.** `action_save` reads budgets via `self._input(f"fork-config-budget-{i}")` for `i in range(count)`, and `_input` does an unguarded `query_one` that raises `NoMatches`. The `_maybe_resize_rows` guard only covers count changes, not the async mount race for an individual row — so a row that hasn't finished mounting aborts Save with an unhandled traceback instead of a friendly error.

Suggest: guard the lookups (`query` + presence check) and surface a validation message.

apps/cli/widgets/fork_tabs.py
```diff
@@ -0,0 +1,141 @@
...
+                self._chip_text(status),
+                classes="fork-tab",
+                id=chip_id,
+            )
+            chip.can_focus = True
+            if status.id == self.active_id:
+                chip.add_class("active")
+            await self.mount(chip)
+
+    def watch_branch_costs(
```
DEENUU1, line 99: 🟡 **Cost render dropped when it races chip mount.** `watch_statuses` mounts status chips with `await self.mount(...)` (async), but `watch_branch_costs` is synchronous and looks chips up via `next(..., None)`. `_poll_fork_state` sets `statuses` then `branch_costs` in the same event-loop turn, so on a freshly (re)mounted chip the cost lookup returns `None` and the `$x/$y` text is silently skipped — persistently lagging a tick with `always_update` churn.

Suggest: render cost inside the same watcher that mounts the chip, or store cost on the chip at mount time.

apps/cli/screens/chat.py
```diff
@@ -384,13 +523,101 @@ def _show_welcome(self) -> None:
...
+    def _is_fork_inspection(text: str) -> bool:
+        """Return True for ``/fork diff`` and its argumented form.
+
+        Inspection commands are allowed during an active fork —
+        but only the ``diff`` sub-command of ``/fork``. ``/fork``
+        without args would re-enter the picker modal and clobber state;
+        ``/fork-config`` is similarly blocked.
+        """
+        stripped = text.strip()
+        return stripped.startswith("/fork ") and stripped[6:].lstrip().startswith("diff")
```
DEENUU1, line 537: 🟡 **`/FORK diff` (uppercase) is wrongly blocked during a fork.** `_is_fork_inspection` tests `text.strip().startswith("/fork ")` on the original-case text, while command dispatch lowercases (`text.split()[0].lower()`). So `/FORK diff` fails the allow-check and is rejected even though `/fork diff` is permitted.

Suggest: lowercase before the `startswith` check.

apps/cli/config.py
```diff
@@ -262,9 +342,26 @@ def _coerce_value(key: str, value: str) -> Any:
...
         return float(value)
     if key in ("shell_allow_list", "approve_tools"):
         return [v.strip() for v in value.split(",") if v.strip()]
+    if key == "fork_branch_models":
+        return [v.strip() or None for v in value.split(",")] if value else []
```
DEENUU1: 🟡 **`fork_branch_models`/`fork_branch_budgets` lose positional length on round-trip.** `_coerce_value` returns `[]` for an empty string (`... if value else []`). A 1-branch all-default config persists these as `""` and reloads as `[]` rather than `[None]`, so the per-branch list length no longer matches `fork_branch_count`. The picker happens to pad, but the persisted/coerced value is positionally wrong and any code indexing model/budget by branch slot relies on that padding to avoid an `IndexError`.

Suggest: coerce to a count-aligned list (pad with `None`) rather than `[]`.

tests/test_forking_core.py
```diff
@@ -0,0 +1,2892 @@
...
+
+
+def test_clone_for_branch_shared_message_queue():
+    parent_queue = MessageQueue()
+    deps = DeepAgentDeps(backend=StateBackend(), message_queue=parent_queue)
+    cloned = clone_for_branch(deps, BranchIsolation(message_queue="shared"))
+    assert cloned.message_queue is parent_queue
+
+
+def test_clone_for_branch_copy_todos():
```
DEENUU1, line 490: 🔴 **This test codifies the `todos="copy"` bug (batch 1).** `test_clone_for_branch_copy_todos` asserts `cloned.todos == []` — but the policy is named/documented "copy", so it should inherit a *copy* of the parent's todos. The test locks in the buggy empty-list behaviour, so the suite would go red when the bug is fixed. This is exactly why the defect ships green.

Fix: seed `deps.todos` with an item, then assert `cloned.todos == deps.todos` **and** `cloned.todos is not deps.todos` (independent copy), and mutate `cloned.todos` to assert the parent's list is unaffected. If empty-on-fork is genuinely intended, rename the policy value to `"empty"`/`"isolated"` so the name stops lying.

tests/test_forking_merge_apply.py
```diff
@@ -0,0 +1,805 @@
...
+    assert result.errors == []
+    assert parent.read_bytes("cat.md") == b"original"
+
+
+# ---------------------------------------------------------------------------
+# Test D — Conflict surfacing (parent modified by a third actor).
+# ---------------------------------------------------------------------------
+
+
+async def test_conflict_surfaced_when_parent_modified_between_fork_and_merge(
```
DEENUU1, line 162: 🔴 **This test codifies the conflict-clobber data-loss bug (batch 2).** `test_conflict_surfaced_when_parent_modified_between_fork_and_merge` asserts the parent ends up with the branch's bytes (`parent.read_bytes('cat.md') == b'alpha wrote cat'`) and only that the path appears in `result.conflicts`. So it cements last-write-wins: a third actor's concurrent edit to the parent is overwritten, with the conflict merely *reported* after the destruction.

Decide the contract: if conflicts should be non-destructive, assert the conflicting path is **not** overwritten (or a backup is written) and change `flush_to` to skip/back-up conflicted paths. If clobber is truly intended, document it explicitly and add a test that a caller can abort *before* commit.

tests/test_tui_forking.py
```diff
@@ -0,0 +1,1722 @@
...
+            msg1 = [ModelResponse(parts=[TextPart(content="hello")])]
+            panel.replay_messages_append(msg1)
+            assert panel._last_replayed_len == 1
+
+            msg2 = msg1 + [ModelResponse(parts=[TextPart(content="world")])]
+            panel.replay_messages_append(msg2)
+            assert panel._last_replayed_len == 2
+            await _drain_tasks(session)
+
+    async def test_e1b_b_tool_call_then_return_updates_in_place(self, fork_app: DeepApp) -> None:
```
DEENUU1, line 1603: 🟠 **Incremental-replay test is constructed to dodge the stuck-tool bug (batch 2).** `test_e1b_b_tool_call_then_return_updates_in_place` puts the `ToolCallPart` and `ToolReturnPart` on a tick boundary with **no intervening `TextPart`**, and asserts only `_last_replayed_len == 2` / `'c1' in _rendered_call_ids` — never that the call actually *completed*. The real defect needs a slice that ends in a `TextPart` (which sets `_current_assistant=None`), so the next tick's `ToolReturnPart` hits the `if assistant_msg is not None` guard and is dropped, leaving the call spinning forever.

Add a test: tick N = `[ToolCallPart, TextPart]`, tick N+1 = `[ToolReturnPart(same id)]`; assert the tool call renders as completed with its return content.

tests/test_tui_forking.py
```diff
@@ -0,0 +1,1722 @@
...
+            assert fork_app.active_fork is session
+            panels = list(fork_app.screen.query(BranchPanelWidget))
+            assert len(panels) == 2
+            labels = {p.label for p in panels}
+            assert labels == {"a", "b"}
+
+            await _drain_tasks(session)
+
+
+class TestBranchSteering:
```
DEENUU1, line 162: 🟠 **Steering tests only use single-token labels, hiding the `>>label` regex bug (batch 2).** Every `TestBranchSteering` case uses labels like `a`/`b`/`nosuchbranch`. The production regex `^>>(\w+)\s+(.*)` can't match a hyphenated/spaced label (`approach-a`), which the fork picker happily creates — `\w+` captures `approach`, the steer is rejected as "unknown branch".

Add a test that starts a fork with label `approach-a` and posts `>>approach-a focus on tests`; assert it routes to that branch's queue (pending count increments) rather than being rejected.

tests/test_forking_diff.py
```diff
@@ -0,0 +1,821 @@
...
+    assert pd.branches["b"].operation == "untouched"
+    assert report.summary.per_branch_unique == {"a": 1, "b": 0}
+
+
+# ---------------------------------------------------------------------------
+# 4. Binary file → is_binary=True, placeholder content, no unified diff
+# ---------------------------------------------------------------------------
+
+
+async def test_diff_binary_file_returns_placeholder() -> None:
```
DEENUU1, line 177: 🟠 **No test for divergent binary writes — masks the binary misclassification bug (batch 2).** The only binary test (`test_diff_binary_file_returns_placeholder`) uses a single branch. `_classify_agreement` keys on `(operation, new_content, is_binary)` and `new_content` is always `None` for binary, so two branches writing **different** binary bytes to the same path are labelled `unanimous_change` instead of `split` (verified empirically: `b'\x00\x01\x02'` vs `b'\x00\x09\x09'` → `unanimous_change`, `split_paths=0`).

Add a test: two branches write different binary bytes to the same path → assert `agreement == 'split'`. Fix by comparing a content hash for binary changes.

tests/test_forking_core.py
```diff
@@ -0,0 +1,2892 @@
...
+    rt = next(iter(coord.branches.values()))
+    assert rt.task.cancelled()
+
+
+# ---------------------------------------------------------------------------
+# merge_or_select fails fast when winner task was already cancelled
+# ---------------------------------------------------------------------------
+
+
+async def test_merge_or_select_raises_if_winner_cancelled():
```
DEENUU1, line 942: 🟠 **The crash is asserted as the contract; the graceful path and the failed-winner case are untested.** `test_merge_or_select_raises_if_winner_cancelled` does `pytest.raises(RuntimeError)` — treating the uncaught raise as correct. There is **no** test that picks a *failed* (exception) winner, where `await winner.task` re-raises the branch's own exception raw out of `merge_or_select` (not a `CancelledError`, not caught) — the exact path that crashes `agent.run()` via the tool's `except ValueError` (batch 1).

Add a test forking with a failing agent, let the branch reach `failed`, then `merge_or_select(pick:<id>)` and assert a *graceful* outcome (clean result or a caught/wrapped error string), not an uncaught exception.

tests/test_forking_core.py
```diff
@@ -0,0 +1,2892 @@
...
+    assert rt.status.state == "failed"
+    assert rt.status.error and "boom" in rt.status.error
+
+
+# ---------------------------------------------------------------------------
+# Coordinator aclose cancels outstanding tasks
+# ---------------------------------------------------------------------------
+
+
+async def test_coordinator_aclose_cancels_outstanding_tasks():
```
DEENUU1, line 871: 🟠 **`aclose` test only checks `cancelled()`, not the leak it can cause (batch 1).** `test_coordinator_aclose_cancels_outstanding_tasks` asserts `rt.task.cancelled()` after `sleep(0)`, but never that `aclose` *awaited* the cancelled tasks before `materializer.cleanup()` (`rmtree`). The bug is precisely that it doesn't await — a branch mid-write can recreate files after the rmtree, leaking the fork dir.

Use a slow agent that writes into the branch dir and blocks; call `aclose`; assert (a) tasks are `done()` (not merely cancelled) before cleanup runs, and (b) the fork directory is fully removed with no leftover files.

tests/test_forking_materializer.py
```diff
@@ -0,0 +1,286 @@
...
+    with caplog.at_level(logging.WARNING, logger="pydantic_deep.toolsets.forking.isolation"):
+        result = overlay.write("a.py", "after")
+
+    assert result.error is None
+    assert overlay.read_bytes("a.py") == b"after"
+    assert parent.read_bytes("a.py") == b"before"  # parent untouched
+    assert any("flush_change failed" in record.message for record in caplog.records)
+
+
+def test_safe_relative_strips_leading_slash(tmp_path: Path) -> None:
```
DEENUU1, line 274: 🟠 **`_safe_relative` test only covers a leading slash — no `..` traversal case (batch 1).** `test_safe_relative_strips_leading_slash` exercises `path.lstrip('/')` but never a path containing `..`, which escapes the materializer root (`../../etc/...`).

Add a test passing `'../../etc/passwd'` (and `'a/../../b'`) to `branch_path`/`parent_path` and assert the result `.is_relative_to(root)`. It fails today; the fix is to resolve+clamp under root in `_safe_relative`.

tests/test_forking_budget.py
```diff
@@ -0,0 +1,686 @@
...
+        await tracker.on_cost_update(_cost_info(total=total))
+
+    for bid in handle.branches:
+        assert coord.branches[bid].status.state == "aggregate_budget_exhausted"
+        assert "aggregate budget exhausted" in (coord.branches[bid].status.error or "")
+
+    await _drain_branch_tasks(coord)
+
+
+async def test_fork_cost_returns_correct_breakdown_and_aggregate():
```
DEENUU1, line 156: 🟡 **Budget tests round-trip private state and never drive the self-cancel path (batch 2).** `test_fork_cost_returns_correct_breakdown_and_aggregate` sets `tracker._total_cost_usd` directly and asserts `fork_cost()` echoes it — testing that arithmetic reads a field you just wrote, not that cost is actually accrued. More importantly, every budget test calls `on_cost_update(...)` from the *test* coroutine, never from inside a running branch task — so the bug where the watcher self-cancels its own task (a no-op until the next await) is never exercised, and no test asserts the branch task actually *stopped*.

Drive cost via the public `on_cost_update` path, and add a test triggering the budget watcher from within a long-running branch task that asserts `rt.task.cancelled()` and that the agent did not run to completion.

tests/test_forking_diff.py
```diff
@@ -0,0 +1,821 @@
...
+        "fork-6",
+        [await _make_runtime(branch_id="a", label="alpha", overlay=overlay_a)],
+        paths_filter=["keep.py"],
+    )
+
+    reported_paths = [pd.path for pd in report.paths]
+    assert reported_paths == ["keep.py"]
+
+
+async def test_diff_paths_filter_keeps_untouched_paths_for_transparency() -> None:
```
DEENUU1, line 296: 🟡 **`agreement_score` under a `paths_filter` is never asserted — and it's misleading.** `test_diff_paths_filter_keeps_untouched_paths_for_transparency` comments "keeps agreement_score meaningful" but asserts nothing about the score. The numerator (`split_paths`) is counted over the *filtered* set while the denominator (`total_paths_touched`) is the *full* union — so filtering out a genuinely-conflicting path yields `agreement_score == 1.0` ("perfect agreement") while the branches actually conflict on the hidden path (verified empirically).

Add a test that filters out a real split path and asserts the score is **not** falsely 1.0. Fix by scoping numerator and denominator consistently.

tests/test_forking_judge.py
```diff
@@ -0,0 +1,1676 @@
...
+    judge = JudgeAgent(tm)
+    verdict, usage = await judge.evaluate("goal", _make_report(), _make_outcomes())
+    assert isinstance(verdict, JudgeVerdict)
+    assert verdict.winner_branch_id == "a"
+    assert verdict.confidence == 0.9
+    assert "smaller" in verdict.reasoning
+    assert verdict.rejected_with_reasons == {"b": "B is too verbose."}
+    assert verdict.caveats == ["No tests were run."]
+    # usage is whatever TestModel provides; just confirm it's not None for valid runs
+    assert usage is not None
```
DEENUU1, line 170: 🟡 **Weak assertion + untested vote-path usage shape.** `assert usage is not None` passes for literally any object, so it pins nothing about the judge's token/cost accounting. (Note: `result.usage` *is* a usable `RunUsage` here — not a bug — but the assertion doesn't verify that.) Separately, the single-judge path returns a scalar usage while the vote path (`_run_judges`) returns a **list** of per-judge usages — that type divergence is never asserted (tests only ever see `judge_usage = None`).

Assert `usage.total_tokens > 0` for the single judge, and assert the vote path yields a list of the expected length.

tests/test_tui_forking.py
```diff
@@ -0,0 +1,1722 @@
...
+                await dispatch_command(fork_app, "/fork diff")
+                await pilot.pause()
+
+            # Dispatcher pushes the picker modal — editor invocation happens on user confirm.
+            assert isinstance(fork_app.screen, DiffPickerModal)
+            await fork_app.pop_screen()
+
+            await session.abort()
+
+    async def test_open_diff_falls_back_to_tui_modal(self, fork_app: DeepApp) -> None:
```
DEENUU1, line 1047: 🟡 **`/fork diff` fallback test only checks the modal exists, not that it's browse-mode — masks the silent no-op (batch 2).** `test_open_diff_falls_back_to_tui_modal` asserts a `MergePickerModal` is on the stack and dismisses with `None`; it never presses Enter / picks, and never asserts `view_only`/browse mode. In production the fallback pushes the picker in **pick mode with no callback**, so a user's Enter silently does nothing.

Assert the fallback modal is `view_only` (browse title) **and** that Enter does not commit a merge (`active_fork` unchanged).

tests/test_tui_forking.py
```diff
@@ -0,0 +1,1722 @@
...
+            await pilot.pause()
+            modal = MergePickerModal(report, statuses, {"a": "id-a", "b": "id-b"})
+            app.push_screen(modal)
+            await pilot.pause()
+            modal.dismiss(MergePickerResult(branch_id="id-a"))
+            await pilot.pause()
+
+
+class TestForkPickerValidation:
+    async def test_picker_rejects_blank_or_duplicate_labels(self, fork_app: DeepApp) -> None:
```
DEENUU1, line 719: 🟡 **Duplicate-label test asserts nothing.** `test_picker_rejects_blank_or_duplicate_labels` sets two rows to `x`, calls `action_submit()`, then only `await pilot.pause()` — zero assertions on the rejection. The duplicate-rejection branch is effectively untested (and relatedly, the agent-facing `fork_run` doesn't enforce label uniqueness at all — batch 2).

Assert the error Static shows "distinct" and `app.screen is modal` (not dismissed), mirroring `test_distinct_labels_at_n4` in `test_fork_picker_n_branches.py`.

tests/test_tui_forking.py
```diff
@@ -0,0 +1,1722 @@
...
+
+            await _drain_tasks(session)
+            a_id = session.label_to_id["a"]
+            await session.merge(a_id)
+            fork_app.active_fork = None
+            await pilot.pause()
+            assert not badge.has_class("visible")
+
+
+class TestTabCyclesFocus:
```
DEENUU1, line 398: 🟡 **No test sets branch costs in the same tick as statuses — masks the chip mount race (batch 2).** `TestTabCyclesFocus` is the only place `ForkTabsWidget` is rendered, and it only sets `tabs.statuses`. Because `watch_statuses` mounts chips with `await self.mount(...)` (async) while `watch_branch_costs` is sync, setting both back-to-back (as `_poll_fork_state` does) drops the cost render on the first tick.

Add a test that sets `tabs.statuses` and `tabs.branch_costs` back-to-back, pumps the loop, and asserts the chip text includes the `$x.xx` cost.

DEENUU1 (commented):
## Automated review — batch 3 (test quality)

Reviewed all 18 test files (+12k lines). The suite is broad and the happy paths are well covered, but several tests **lock in the confirmed bugs from batches 1–2** — which is why CI is green at "100% coverage" while the defects are live. Coverage measures lines executed, not assertions that would fail when behavior is wrong. The most valuable fixes below are the tests that *codify* a bug (they'll need to change when the bug is fixed).

**14 inline findings below**, prioritized: tests that mask confirmed bugs first, then tautological/weak tests, then gaps.

**Summary of lower-priority items (not inline):**
- `tests/test_cli_agent.py` — ~14 tests end in `assert agent is not None` with docstrings claiming properties they never check (browser capability presence, working_dir in instructions, lean-mode behaviour). `test_include_browser_import_error_warns` has *no* assertion at all (its comment admits it). Assert the claimed property or collapse to one smoke test.
- `interrupt_on` tests (`tests/test_agent*.py`) — all `assert agent is not None`; none pin `require_execute_approval`/`require_write_approval`, so the True→False flip (batch 1) is untested in either direction. Add an introspection test on the resolved console toolset flags.
- `ForkConfigModal` (`apps/cli/modals/fork_config.py`) — **zero tests** anywhere. The unguarded `query_one` budget-row reads (batch 2) and resize/save paths are entirely uncovered.
- `tests/test_forking_editor.py` — covers the `tui` fallback (good) but never the `env is None` / real-`os.environ` branch; path assertions use POSIX literals (`/parent/cat.md`) → would fail on Windows. Derive expected paths via `str(Path(...))`.
- `tests/test_forking_materializer.py:65` — `test_overlay_write_mirrors_to_disk_within_100ms` asserts wall-clock `< 100ms` → flaky on loaded CI; keep the correctness asserts, drop the timing gate.
- `coordinator.py:656` — the deferred-approval + test-runner path is `# pragma: no cover`'d despite being a real feature; add a runner+approval test rather than excluding it.
- `JudgeVerdict.confidence` — add `Field(ge=0, le=1)` + a rejection test (ties to the unbounded-confidence finding in batch 2).

---
_Generated with Claude Code._

DEENUU1 (approved):

PR #124, opened 2026-05-28
fix: update command

### LINKED ISSUE #122: update command doesnt work
### Description

tried to update the app, doesnt work, still showing 3.19, have to remove first then install fresh

Image

Image

### Steps to Reproduce

just run the update command on the powershell

### Expected Behaviour

update run smoothly

### Actual Behaviour

update is not detected properly

### pydantic-deep Version

0.3.19 (now already updated manually to 0.3.22)

### Python Version

3.13.13

### Operating System

Windows 11

### Additional Context

_No response_

OchnikBartek:
## Summary

`run_update()` used `uv tool upgrade pydantic-deep` whenever `uv` was found in PATH — but this only works if pydantic-deep was installed via `uv tool install`. Users who installed via `pip install pydantic-deep[cli]` would see the upgrade silently fail, with no fallback attempted. Fixes #122.

## Changes

- In `run_update()`: if `uv tool upgrade pydantic-deep` returns a non-zero exit code, fall through to `pip install --upgrade pydantic-deep[cli]` instead of returning the error immediately

## Testing

- [ ] New tests added for any new functionality (required — `make test` enforces 100% coverage)
- [ ] Security scan passes: `make security`

Tested the fallback path manually via mocked subprocess — confirmed `uv tool upgrade` failure correctly triggers pip fallback and returns exit code 0.

## Related Issues

Fixes #122

## Notes for Reviewers

The fallback logic is intentionally sequential: try uv tool upgrade first (correct path for uv-managed installs), then pip (correct path for pip-managed installs). Both paths were already present in the code — the fix just removes the early `return` that prevented the fallback from ever being reached.


DEENUU1 (commented):
## Code review — PR #124

Reviewed both changed files in full and traced the eager import chain (`import pydantic_deep` → `pydantic_deep/toolsets/__init__.py` → `liteparse.py`) plus the existing tests. Inline comments below. Severity: 🔴 high · 🟠 medium · 🟡 low.

### Additional finding (file not in diff, so no inline anchor)

🟡 **`tests/test_cli_update.py:237` — the actual fix path is untested.** The PR's whole point is *uv fails → pip fallback succeeds*, but no test covers it. `test_propagates_nonzero_exit_code` patches `subprocess.run` to return `returncode=1` for **both** calls, so `run_update() == 1` passes regardless and never asserts that pip is invoked after uv fails. Branch coverage stays at 100% (the `return 0` line is hit by `test_uses_uv_when_available`, the fall-through by this test), which masks the gap. Add a test: `_find_uv` returns a path, first `subprocess.run` (uv) returns `1`, second (pip) returns `0`, assert result is `0` **and** that pip was called with `sys.executable ... pydantic-deep[cli]`. The PR's "new tests added" checkbox is correctly left unchecked — this is why.

pydantic_deep/toolsets/liteparse.py
```diff
@@ -11,27 +11,31 @@
...
+# stubs happen to be present.
 _HAS_LITEPARSE = False
+_LiteParse: Any = None
+_CLINotFoundError: type[BaseException] = type("_CLINotFoundError", (Exception,), {})
 
 try:
-    from liteparse import LiteParse as _LiteParse
-    from liteparse.types import CLINotFoundError as _CLINotFoundError
-
+    _LiteParse = importlib.import_module("liteparse").LiteParse
```
DEENUU1, line 36: 🔴 **Uncaught `AttributeError` crashes `import pydantic_deep`.**

`importlib.import_module("liteparse").LiteParse` raises **`AttributeError`** if `liteparse` is installed but lacks the `LiteParse` attribute — and `except ImportError` does **not** catch that. The old `from liteparse import LiteParse` raised `ImportError` (which *was* caught), so this is a regression.

This is exactly the scenario the new comment above anticipates ("its API shape differs between versions"), and because this module is imported eagerly by `toolsets/__init__.py` and `pydantic_deep/__init__.py:166`, the failure takes down the entire `import pydantic_deep` instead of degrading gracefully. Note the sibling `_CLINotFoundError` line below correctly does `contextlib.suppress(ImportError, AttributeError)` — this line should be just as defensive:

```python
with contextlib.suppress(ImportError, AttributeError):
    _LiteParse = importlib.import_module("liteparse").LiteParse
    _HAS_LITEPARSE = True
```

(Keep `_HAS_LITEPARSE = True` inside the `with`, after the assignment, so a failed attribute lookup leaves it `False`.)

pydantic_deep/toolsets/liteparse.py
```diff
@@ -11,27 +11,31 @@
...
 except ImportError:  # pragma: no cover
+    pass
 
-    class _LiteParse:  # type: ignore[no-redef]
-        def __init__(self, **kwargs: Any) -> None: ...  # pragma: no cover
-
-    class _CLINotFoundError(Exception):  # type: ignore[no-redef]
-        pass  # pragma: no cover
+with contextlib.suppress(ImportError, AttributeError):
+    _CLINotFoundError = importlib.import_module("liteparse.types").CLINotFoundError
```
DEENUU1: 🟠 **Decoupling `_CLINotFoundError` from the main import lets the real exception slip past the `except`.**

Previously both imports were inside one `try`, so they succeeded or failed together. Now `liteparse` can import successfully (`_HAS_LITEPARSE = True`) while this line silently fails (AttributeError/ImportError suppressed), leaving `_CLINotFoundError` as the **synthetic placeholder** from line 29.

At runtime, liteparse then raises its *real* `CLINotFoundError`, which is a different class — so `except _CLINotFoundError` at lines 141/195 won't match, and the error falls through to the generic `except Exception` returning `"Parse error: …"` instead of the intended `"LiteParse CLI not found: …"`. The same placeholder also leaks out via the public `LiteparseCliNotFoundError` export (line 207), so downstream `except LiteparseCliNotFoundError` would miss the real one too.

Worth gating on `_HAS_LITEPARSE` and/or logging when the type can't be resolved, rather than silently substituting a class that can never match what liteparse raises.

apps/cli/update.py
```diff
@@ -113,7 +113,9 @@ def run_update() -> int:
     """
     uv = _find_uv()
     if uv:
-        return subprocess.run([uv, "tool", "upgrade", "pydantic-deep"]).returncode
+        result = subprocess.run([uv, "tool", "upgrade", "pydantic-deep"])
```
DEENUU1: 🟠 **Fallback fires on *any* non-zero uv exit, not just "not a uv install".**

The fix can't distinguish "pydantic-deep wasn't installed via `uv tool`" (the #122 case, where falling back to pip is correct) from a transient uv failure (network, lock, registry hiccup) on a genuinely uv-managed install. In the latter case the fallback runs `pip install --upgrade` against `sys.executable` — which for a `uv tool`-managed install points at uv's **isolated tool venv that typically has no `pip`** — producing a confusing secondary failure that masks the real uv error and returns pip's exit code.

Consider scoping the fallback to the actual #122 signal (e.g. inspect uv's output for "is not installed", or only fall back when the package isn't a uv-managed tool) so a transient uv failure on a uv install surfaces the real error instead of a misleading pip one.

PR #125, opened 2026-05-28
feat: automatic fallback-model retry on primary model failure

### LINKED ISSUE #112: Feature: automatic fallback-model retry on primary model failure
## Context

When the primary model fails (rate limit, 5xx, timeout, provider outage, context-window overflow), `agent.run()` raises and the whole run dies. Pydantic-AI already ships `FallbackModel` — but users must wire it up by hand for every agent.

## Proposal

Expose a first-class `fallback_model` arg on `create_deep_agent()` that wraps the primary in a `FallbackModel` automatically.

```python
from pydantic_deep import create_deep_agent

agent = create_deep_agent(
    model="anthropic:claude-sonnet-4-6",
    fallback_model="anthropic:claude-haiku-4-5",  # str | Model | list[str|Model]
)
```

Accept:
- `str` — single fallback
- `list[str | Model]` — ordered fallback chain
- `None` (default) — opt-in, no change in behavior

## Retry semantics

- Retry on: `ModelHTTPError` (429, 5xx), `UnexpectedModelBehavior`, provider timeout
- Do NOT retry on: `ModelRetry` (tool-driven), auth errors (401/403), validation errors
- Preserve message history across fallback hop — fallback continues conversation, not restart
- Emit hook event `MODEL_FALLBACK_TRIGGERED` (new) with `{primary, fallback, error}` payload for observability

## Default fallback chain (opt-in via flag)

```python
create_deep_agent(model="...", use_default_fallbacks=True)
```

Built-in chain per provider family:
- Anthropic: `sonnet → haiku`
- OpenAI: `gpt-5 → gpt-5-mini`
- Bedrock/Vertex: same-family smaller variant

## Why now

- Production runs die on transient 429s — Anthropic capacity events in past months
- `HooksCapability` + `RUN_ERROR` already catch failures but can't resume mid-run
- Lowers ops burden vs. external retry/circuit-breaker layers

## Acceptance

- [ ] `fallback_model` kwarg on `create_deep_agent()`
- [ ] `use_default_fallbacks` flag wiring built-in chains
- [ ] `MODEL_FALLBACK_TRIGGERED` hook event
- [ ] Tests: forced primary failure → fallback completes run with intact message history
- [ ] Docs page `advanced/fallback-models.md`

OchnikBartek:
## Summary

Adds opt-in automatic fallback to a secondary model when the primary fails with a transient API error (rate limit, 5xx, provider outage). Exposed both as a library parameter on `create_deep_agent()` and as a UI flow in the TUI (`/model` → optional fallback picker → persisted to config).

## Changes

- **Library** — new `fallback_model: str | Model | list[str | Model] | None = None` parameter on `create_deep_agent()`. Wraps the primary model in pydantic-ai's `FallbackModel`. Accepts a single model or an ordered chain.
- **Auth-error filtering** — fallback fires on `ModelAPIError` minus messages containing `401`, `403`, `unauthorized`, `forbidden` (case-insensitive). Auth failures propagate instead of silently hopping to the next model where they would fail identically.
- **New hook event** — `HookEvent.MODEL_FALLBACK_TRIGGERED` + `HooksCapability.dispatch_model_fallback()` for observability. Payload carries `primary`, `fallback` (first in chain), and the original exception.
- **TUI** — `/model` now chains into a new `FallbackPickerModal` that lists candidate fallbacks (provider/key status indicators included) plus a "No fallback" option. Selection is persisted to `.pydantic-deep/config.toml`.
- **CLI plumbing** — `CliConfig` gained `fallback_model: str | None`. `DeepApp.reconfigure_agent(model)` simplified to one parameter; new `DeepApp.set_fallback_and_reconfigure(model, fallback)` helper persists the fallback first, then reconfigures (eliminates a sentinel-default-arg pattern).
- **Test isolation** — new autouse `isolate_cli_config` fixture in `tests/conftest.py` redirects `DEFAULT_CONFIG_PATH` to a temp file so tests don't pick up the developer's local `.pydantic-deep/config.toml`.
- **Docs** — new `docs/advanced/fallback-models.md` covering basic usage, CLI flow, exact retry semantics (auth-error substring list spelled out), hook payload, and an explicit "Limitations and caveats" section (static `fallback` field, cost accumulation, substring-based detection, `agent.model` returns the wrapper).
- **Cleanup** — removed dead `except (NoMatches, Exception)` and `contextlib.suppress(NoMatches, Exception)` patterns in `apps/cli/app.py` (`NoMatches` is unreachable when `Exception` is in the same handler).

## Testing

`TestFallbackModel` covers: single-model wrapping, list-chain wrapping, `None` does not wrap, hook dispatched on `ModelAPIError`, hook **not** dispatched on non-`ModelAPIError`, hook **not** dispatched on auth-error messages (`401 unauthorized`, `403 forbidden`, `unauthorized access`).

1573 tests pass, 100% coverage, pyright + mypy clean.

## Related Issues

Closes #112


pydantic_deep/agent.py
```diff
@@ -531,6 +575,19 @@ class CodeAnalysis(BaseModel):
...
+        _fallbacks: list[str | Model] = (
+            list(fallback_model) if isinstance(fallback_model, list) else [fallback_model]
+        )
+        _fallback_hooks = [
+            h for h in (hooks or []) if h.event == HookEvent.MODEL_FALLBACK_TRIGGERED
+        ]
+        if _fallback_hooks:
+            model = _wrap_with_fallback_and_hooks(model, _fallbacks, _fallback_hooks, backend)
+        else:
+            model = FallbackModel(model, *_fallbacks)
```
DEENUU1: **Auth-error filtering is silently bypassed unless a `MODEL_FALLBACK_TRIGGERED` hook is registered — this defeats the feature's documented guarantee.**

The custom `_fallback_on` (which excludes `401/403/unauthorized/forbidden`) is only installed via `_wrap_with_fallback_and_hooks` in the `if _fallback_hooks:` branch. This `else` branch builds a plain `FallbackModel(model, *_fallbacks)`, which uses pydantic-ai's **default** `fallback_on=(ModelAPIError,)`.

`ModelHTTPError` (the 401/403 type) subclasses `ModelAPIError`, so in the default/common case (no observability hook) a `401`/`403` **does** trigger fallback — exactly what the PR summary and `docs/advanced/fallback-models.md` ("Auth/permission errors (401/403)… The original exception is re-raised") promise will *not* happen. The auth-suppression behavior should not depend on whether someone is observing it.

Suggest always installing an auth-filtering `fallback_on`, and making only the hook dispatch conditional, e.g.:

```python
if fallback_model is not None:
    _fallbacks: list[str | Model] = (
        list(fallback_model) if isinstance(fallback_model, list) else [fallback_model]
    )
    _fallback_hooks = [h for h in (hooks or []) if h.event == HookEvent.MODEL_FALLBACK_TRIGGERED]
    # always filters auth errors; dispatches hooks only when present
    model = _wrap_with_fallback_and_hooks(model, _fallbacks, _fallback_hooks, backend)
```

(then have `_fallback_on` skip the dispatch when `fallback_hooks` is empty). A regression test that asserts `_fallback_on` returns `False` for a `401` ModelHTTPError **without** any hook registered would lock this down.

DEENUU1 (commented):
Review of the fallback-model feature. The overall design (wrapping in pydantic-ai's `FallbackModel`, hook event, TUI flow, config isolation fixture) is solid and well-tested. Found one correctness bug that defeats a documented guarantee, plus a few smaller issues — inline below, most severe first.

Minor note not worth an inline comment: `tests/test_agent_extended.py` exercises the fallback decision via the private `agent.model._exception_handlers[0]` attribute of pydantic-ai's `FallbackModel`. The code comments acknowledge this, but it will silently break on a pydantic-ai refactor — worth a follow-up if a public hook ever lands.

Also note this PR bundles the unrelated `liteparse.py` import refactor (same change as #126); the two liteparse comments below apply here too.

docs/advanced/fallback-models.md
```diff
@@ -0,0 +1,93 @@
...
+)
+```
+
+The `HookInput` payload for this event:
+
+| Field | Value |
+|-------|-------|
+| `tool_input["primary"]` | Name of the primary model |
+| `tool_input["fallback"]` | Name of the **first** fallback in the chain (see limitation below) |
+| `tool_error` | The original `ModelAPIError` instance |
```
DEENUU1: **Inaccurate: `tool_error` is not the exception instance — it's a string.**

`_build_hook_input` stores `tool_error=str(tool_error) if tool_error is not None else None` (hooks.py:160), and `HookInput.tool_error` is typed `str | None` (hooks.py:85). So a user following this table and doing `inp.tool_error.status_code` or `isinstance(inp.tool_error, ModelAPIError)` will get an `AttributeError` / always-`False`.

The limitations section (line 84, "inspect `tool_error`…") reinforces the same misconception. Either change the row to `The stringified original error message` (matching the `print(... {inp.tool_error})` example on line 63, which works fine), or — if access to the real exception object is intended — pass it through unstringified.

pydantic_deep/agent.py
```diff
@@ -42,6 +44,40 @@
...
+    first_fallback = fallbacks[0]
+    first_fallback_name = first_fallback if isinstance(first_fallback, str) else str(first_fallback)
+
+    async def _fallback_on(exc: Exception) -> bool:
+        if not isinstance(exc, ModelAPIError):
+            return False
+        message = str(exc).lower()
+        if any(marker in message for marker in _AUTH_ERROR_MARKERS):
+            return False
+        await hooks_cap.dispatch_model_fallback(primary_name, first_fallback_name, exc, backend)
```
DEENUU1: **The hook fires even when the fallback chain is exhausted (no model left to fall back to).**

`FallbackModel.request` calls `_should_fallback(exc)` for *every* model that raises, including the last one in the chain — when it returns `True` the loop simply ends and raises `FallbackExceptionGroup`. Since `_fallback_on` dispatches `MODEL_FALLBACK_TRIGGERED` before returning `True`, a chain `[primary, A]` where `A` also fails emits a "fallback triggered" event for `A`'s failure even though nothing more is tried. Observability consumers will over-count fallbacks and see a hop to a model that never actually ran. Consider only dispatching when a subsequent model exists (the handler can't see the index, so this may need to be tracked via a counter or handled at the `FallbackModel` boundary).

pydantic_deep/toolsets/liteparse.py
```diff
@@ -11,27 +11,31 @@
...
+# stubs happen to be present.
 _HAS_LITEPARSE = False
+_LiteParse: Any = None
+_CLINotFoundError: type[BaseException] = type("_CLINotFoundError", (Exception,), {})
 
 try:
-    from liteparse import LiteParse as _LiteParse
-    from liteparse.types import CLINotFoundError as _CLINotFoundError
-
+    _LiteParse = importlib.import_module("liteparse").LiteParse
```
DEENUU1, line 36: **Regression: `AttributeError` is not caught, so a `liteparse` version skew crashes module import.**

`importlib.import_module("liteparse")` succeeds whenever the package is installed; the `.LiteParse` access then raises `AttributeError` if that version doesn't expose `LiteParse`. `except ImportError` does not catch `AttributeError`, so it propagates and `import pydantic_deep.toolsets.liteparse` hard-crashes. The old `from liteparse import LiteParse` raised `ImportError` (caught) and degraded gracefully. Note line 37 below correctly does `contextlib.suppress(ImportError, AttributeError)` — apply the same here:

```python
except (ImportError, AttributeError):  # pragma: no cover
    pass
```

pydantic_deep/toolsets/liteparse.py
```diff
@@ -11,27 +11,31 @@
...
     _HAS_LITEPARSE = True
```
DEENUU1, line 37: **`_HAS_LITEPARSE` is now set `True` even when the `CLINotFoundError` import fails**, which degrades error reporting.

Previously both imports shared one `try`, so the flag required *both* to succeed. Now if `liteparse` is installed but `liteparse.types.CLINotFoundError` is missing (version skew), `_HAS_LITEPARSE` is `True` while `_CLINotFoundError` stays the synthesized dummy (line 29). The `except _CLINotFoundError` handlers then catch a class that is never raised, so a real CLI-not-found error falls through to the generic `except Exception` and is reported as `Parse error:` instead of `LiteParse CLI not found:`. The exported `LiteparseCliNotFoundError` alias is also the dummy, so user `except LiteparseCliNotFoundError` won't match.

PR #126, opened 2026-05-28
feat: additional built-in output styles

### LINKED ISSUE #111: Enhancement: ship additional built-in output styles
## Context

Built-in [output styles](https://vstorm-co.github.io/pydantic-deepagents/advanced/output-styles/) (`concise`, `explanatory`, `formal`, `conversational`) cover conversational shapes well. Missing: machine-readable / pipe-friendly shapes that downstream tooling expects.

## Proposal

Add these built-ins to `BUILTIN_STYLES`:

| Style | Use case |
|-------|----------|
| `markdown` | Forces well-formed markdown with headings, fenced code blocks, no prose-only replies |
| `json-only` | Output strictly valid JSON; no preamble, no fences, no commentary |
| `bullet` | All answers as bulleted lists; no paragraphs |

## Why

- `json-only` enables piping agent output into jq / downstream parsers without custom system prompts
- `markdown` ensures stable rendering when output is fed into doc generators / UIs
- `bullet` is a common ask for brief status / planning outputs

## Acceptance

- [ ] Three styles added to `BUILTIN_STYLES` registry
- [ ] Docs table in `advanced/output-styles.md` updated
- [ ] Tests verifying `resolve_style("json-only")` etc. return expected `OutputStyle`

OchnikBartek:
## Summary

Adds three new built-in output styles (`markdown`, `json-only`, `bullet`) to cover machine-readable / pipe-friendly response shapes that the existing conversational styles (concise / explanatory / formal / conversational) don't address.

## Changes

- Added `MARKDOWN_STYLE`, `JSON_ONLY_STYLE`, `BULLET_STYLE` constants in `pydantic_deep/styles.py` and registered them in `BUILTIN_STYLES`
- Updated module docstring to list the new styles
- Extended `docs/advanced/output-styles.md`: new rows in the built-ins table, usage examples for each, comparison section explaining when `json-only` is the right call vs `output_type` (structured output), updated resolution order and components count
- Added tests in `tests/test_styles.py` covering each new style constant, registry entry, and `resolve_style()` lookup (including the hyphenated `json-only` key)

## Testing

- [x] All tests pass locally: `make test` — 1559 passed, 100% coverage
- [x] Type checking passes: `make typecheck` — pyright 0 errors, mypy clean

## Related Issues

Closes #111

## Notes for Reviewers

The `json-only` style overlaps conceptually with `output_type=SomeModel` (structured output). The docs include a dedicated comparison section steering users toward `output_type` when the shape is known, and listing the legit use cases where `json-only` is the right choice (free-form JSON, providers without structured-output support, CLI piping, streaming, intermediate tool outputs). The style itself is a prompt directive only — no validation guarantees, by design.


pydantic_deep/toolsets/liteparse.py
```diff
@@ -11,27 +11,31 @@
...
+# stubs happen to be present.
 _HAS_LITEPARSE = False
+_LiteParse: Any = None
+_CLINotFoundError: type[BaseException] = type("_CLINotFoundError", (Exception,), {})
 
 try:
-    from liteparse import LiteParse as _LiteParse
-    from liteparse.types import CLINotFoundError as _CLINotFoundError
-
+    _LiteParse = importlib.import_module("liteparse").LiteParse
```
DEENUU1, line 36: **Regression: `AttributeError` is not caught here, so a version skew crashes module import.**

`importlib.import_module("liteparse")` succeeds whenever the package is installed; the `.LiteParse` attribute access then raises **`AttributeError`** if that version of `liteparse` doesn't expose `LiteParse`. `except ImportError` does *not* catch `AttributeError`, so it propagates and `import pydantic_deep.toolsets.liteparse` (and anything importing it) hard-crashes.

The old code used `from liteparse import LiteParse`, which raised `ImportError` for a missing name and was caught — i.e. it degraded gracefully to `_HAS_LITEPARSE = False`. Note the inconsistency with line 37 below, where you correctly do `contextlib.suppress(ImportError, AttributeError)`.

```python
try:
    _LiteParse = importlib.import_module("liteparse").LiteParse
    _HAS_LITEPARSE = True
except (ImportError, AttributeError):  # pragma: no cover
    pass
```

pydantic_deep/toolsets/liteparse.py
```diff
@@ -11,27 +11,31 @@
...
     _HAS_LITEPARSE = True
```
DEENUU1, line 37: **`_HAS_LITEPARSE` is now `True` even when `CLINotFoundError` failed to import**, which silently degrades error reporting.

Previously both imports lived in one `try` block, so `_HAS_LITEPARSE` was only `True` when *both* `LiteParse` and `CLINotFoundError` were available. Now if `liteparse` is installed but `liteparse.types.CLINotFoundError` is missing (version skew), this sets `_HAS_LITEPARSE = True` while `_CLINotFoundError` stays the locally-synthesized dummy class (line 29).

Consequence: the `except _CLINotFoundError` handlers in `parse_document` / `screenshot_document` catch a class that is never raised, so a genuine CLI-not-found error falls through to the generic `except Exception` and is reported as `Parse error: ...` instead of `LiteParse CLI not found: ...`. The exported `LiteparseCliNotFoundError` alias (line 207) is likewise the dummy, so user code doing `except LiteparseCliNotFoundError` won't match the real exception.

docs/advanced/output-styles.md
```diff
@@ -32,8 +35,52 @@ agent = create_deep_agent(output_style="formal")
...
+agent = create_deep_agent(output_style="markdown")
+
+# Strict JSON for piping into jq / parsers
+agent = create_deep_agent(output_style="json-only")
+
+# Bulleted summaries / status updates
+agent = create_deep_agent(output_style="bullet")
 ```
 
+`json-only` is a prompt — not a guarantee"
```
DEENUU1: **Typo + broken rendering.** There is a stray trailing `"` after "not a guarantee", and there's no blank line before the following paragraph, so Markdown merges this line into the next paragraph instead of rendering it as the intended callout/heading.

```markdown
**`json-only` is a prompt — not a guarantee.**

The model is *asked* to produce valid JSON, but nothing validates the
result. ...
```

DEENUU1 (commented):
Code review of the output-styles change.

The styles themselves and their tests look good. Two notes outside the inline comments:

1. **Scope:** this PR is titled "additional built-in output styles" but also carries an unrelated refactor of `pydantic_deep/toolsets/liteparse.py` (the dynamic-import change). Bundling it makes the diff harder to reason about and the regression below easier to miss — consider splitting it into its own PR/commit.
2. The new style constants (`MARKDOWN_STYLE`, `JSON_ONLY_STYLE`, `BULLET_STYLE`) are correct and well-covered.

Inline findings below, most severe first.

PR #127, opened 2026-05-28
feat: security hook preset for tool misuse prevention

### LINKED ISSUE #110: Feature: built-in security hook preset for tool misuse prevention
## Context

The [hooks system](https://vstorm-co.github.io/pydantic-deepagents/advanced/hooks/) exposes `PRE_TOOL_USE` with deny-on-`allow=False` semantics — powerful, but every user must hand-roll their own safety gate. Misuse patterns (rm -rf, fork bombs, path traversal, secret exfiltration via `read_file`, outbound network calls from `execute`) repeat across deployments.

## Proposal

Ship a `SecurityHook` preset (or `pydantic_deep.hooks.security` module) wired via:

```python
from pydantic_deep.hooks import default_security_hook

agent = create_deep_agent(hooks=[default_security_hook()])
```

## Default rules (opt-out, not opt-in)

- `execute`: deny `rm -rf /`, fork bombs (`:(){:|:&};:`), `mkfs`, `dd if=… of=/dev/…`, curl-piped-to-shell
- `write_file` / `edit_file`: block writes outside cwd / sandbox root (path-traversal regex on `..`)
- `read_file`: deny `/etc/shadow`, `~/.ssh/*`, `.env*`, AWS/GCP credential paths
- `POST_TOOL_USE` redactor: scrub `AKIA[0-9A-Z]{16}`, `sk-[A-Za-z0-9]{20,}`, `ghp_*`, JWT-shaped tokens from tool output

## API sketch

```python
default_security_hook(
    blocked_commands: list[str] | None = None,   # extend defaults
    allowed_write_roots: list[Path] | None = None,
    blocked_read_paths: list[str] | None = None,
    redact_secrets: bool = True,
    mode: Literal["deny", "warn"] = "deny",
)
```

## Why now

- Issue #31 (`Command Injection in BackendSkillScriptExecutor`) shows misuse surface is real
- Docs already position hooks as the security primitive; no batteries-included gate exists
- Lowers blast radius for users running `LocalBackend` against untrusted prompts

## Acceptance

- [ ] `default_security_hook()` exported from `pydantic_deep.hooks`
- [ ] Unit tests covering each rule (deny + allow paths)
- [ ] Docs page `advanced/security-hook.md` with enable/extend recipe
- [ ] Example in `examples/security_gate/`

OchnikBartek:
## Summary

Adds `default_security_hook()` — a batteries-included hook preset that blocks common tool-misuse patterns (destructive shell commands, path-traversal writes, sensitive-file reads) and redacts obvious secrets from tool output. Drop-in via `create_deep_agent(hooks=default_security_hook())`.

## Changes

- New `default_security_hook()` factory in `pydantic_deep/capabilities/hooks.py` (returns `list[Hook]`) + three `DEFAULT_*` regex tuples for blocked commands, sensitive read paths, and secret-token shapes
- Exported from `pydantic_deep` top-level package
- `docs/advanced/hooks.md`: added "Built-in Security Preset" section (defaults, customization, shadow mode, limits) and extended the Components table
- `tests/test_hooks.py`: 51 new tests (`TestSecurityHook*` classes) covering every rule, custom-pattern override, warn mode, and `create_deep_agent` integration
- `examples/security_gate/`: runnable demo (`main.py` + `README.md`)

## Testing

- [x] All tests pass locally: `make test` (1604 passed, 100% coverage)
- [x] Type checking passes: `make typecheck` (pyright + mypy clean)

## Related Issues

Closes #110


pydantic_deep/capabilities/hooks.py
```diff
@@ -447,12 +448,250 @@ async def after_model_request(
         return response
 
 
+# Default destructive-command patterns matched against the ``command`` arg of
+# the ``execute`` tool. Each entry is a regex; the first match denies the call.
+DEFAULT_BLOCKED_COMMANDS: tuple[str, ...] = (
+    # ``rm -rf /`` and variants like ``rm -rfv /``, ``rm -fr /``
+    r"\brm\s+-[a-zA-Z]*[rR][a-zA-Z]*[fF][a-zA-Z]*\s+/(?:\s|$)",
```
DEENUU1: 🟠 **The `rm -rf /` patterns catch essentially only the one variant the OS already refuses, and miss every effective one.** The trailing `\s+/(?:\s|$)` requires the argument to be *exactly* `/`. Verified against the compiled defaults:

```
BLOCK 'rm -rf /'
ALLOW 'rm -rf /*'              ← wipes everything, NOT stopped by --preserve-root
ALLOW 'rm -rf /home'
ALLOW 'rm --recursive --force /'  ← long options: -[a-zA-Z]* can't match '--'
ALLOW 'rm -fr ~'               ← home wipe
```

Note that plain `rm -rf /` is refused by GNU coreutils by default (`--preserve-root`), so the headline example the docs table advertises is the *least* useful one to catch, while `rm -rf /*` (equally catastrophic, **not** preserve-root protected) sails through. Consider matching `rm` with recursive+force flags (incl. `--recursive`/`--force` long forms) followed by `/`, `/*`, `~`, `$HOME`, `.`, etc., rather than only a bare `/`.

pydantic_deep/capabilities/hooks.py
```diff
@@ -447,12 +448,250 @@ async def after_model_request(
...
+    r"(?:^|/|~/)\.config/gcloud(?:/|$)",
+    r"application_default_credentials\.json$",
+)
+
+
+# Secret-shaped tokens redacted from POST_TOOL_USE output. Each pattern is
+# replaced with ``[REDACTED]`` when ``redact_secrets`` is enabled.
+DEFAULT_SECRET_PATTERNS: tuple[str, ...] = (
+    r"AKIA[0-9A-Z]{16}",  # AWS access key ID
+    r"sk-[A-Za-z0-9]{20,}",  # OpenAI-style API keys
```
DEENUU1: 🟠 **`sk-[A-Za-z0-9]{20,}` does not match current OpenAI key formats, so they leak unredacted.** Modern keys are `sk-proj-…`, `sk-svcacct-…`, `sk-admin-…` and contain `-`/`_`. The character class `[A-Za-z0-9]` breaks at the first hyphen — `sk-proj` is only 7 chars before the `-`, below the `{20,}` floor — so no match. Verified:

```
REDACT 'sk-abc123def456ghi789jklmno'                    (legacy key)
LEAK   'sk-proj-abc123XYZ_def456-ghi789jklmnopqrstuv'   (current default format)
```

The comment calls this "OpenAI-style API keys," but the format OpenAI has issued by default since 2024 isn't covered. Allow `[-_]` in the body (e.g. `sk-(?:proj|svcacct|admin)?-?[A-Za-z0-9_-]{20,}`) so the current keys are redacted.

pydantic_deep/capabilities/hooks.py
```diff
@@ -447,12 +448,250 @@ async def after_model_request(
...
+    args: dict[str, Any],
+    *,
+    allowed_write_roots: Sequence[str] | None,
+) -> str | None:
+    """Return a denial reason if ``write_file``/``edit_file`` targets an unsafe path."""
+    path = args.get("path")
+    if not isinstance(path, str):
+        return None
+    # Path-traversal segments are always suspicious, even without explicit roots.
+    if re.search(r"(?:^|/)\.\.(?:/|$)", path):
```
DEENUU1, line 584: 🟠 **Write protection is asymmetric — sensitive *writes* are wide open by default.** Without `allowed_write_roots`, the only thing blocked here is `..` traversal. So by default an agent may `write_file("~/.ssh/authorized_keys", …)` (persistence), `write_file("/etc/cron.d/x", …)`, or clobber `~/.bashrc` — all absolute paths with no `..`. Meanwhile the *read* side (`DEFAULT_BLOCKED_READ_PATHS`) carefully blocks `.ssh`, `.aws/credentials`, etc. Writing `~/.ssh/authorized_keys` is arguably a higher-impact vector than reading it. Consider applying a sensitive-path denylist to writes too (the read list is a natural starting point), independent of `allowed_write_roots`.

pydantic_deep/capabilities/hooks.py
```diff
@@ -447,12 +448,250 @@ async def after_model_request(
...
+)
+
+
+# Tools whose ``path`` argument must stay inside ``allowed_write_roots``.
+_WRITE_TOOLS: tuple[str, ...] = ("write_file", "edit_file")
+
+
+def _normalize_path(raw: str) -> str:
+    """Expand ``~`` and resolve a path without requiring it to exist."""
+    return str(Path(raw).expanduser().resolve(strict=False))
```
DEENUU1: 🟠 **Path validation runs in the controller's filesystem namespace, not the agent backend's.** `Path(raw).expanduser().resolve(strict=False)` expands `~` against the *controller* process's `$HOME` and resolves relative paths against the controller's CWD. But the tools operate on `deps.backend`, which for `DockerSandbox`/remote backends is a different filesystem with a different home and CWD. So `allowed_write_roots` enforcement can (a) accept a relative path that escapes the intended root inside the sandbox, or (b) reject a valid one — the `~`/relative resolution simply doesn't refer to the namespace the write will actually hit. The existing tests only use absolute `tmp_path` roots, which masks this. At minimum, document that `allowed_write_roots` assumes absolute backend paths; ideally reject relative paths / `~` rather than resolving them locally.

pydantic_deep/capabilities/hooks.py
```diff
@@ -447,12 +448,250 @@ async def after_model_request(
...
+
+    if redact_secrets:
+        secret_compiled = _compile_all(
+            DEFAULT_SECRET_PATTERNS if secret_patterns is None else secret_patterns
+        )
+
+        async def _post_tool_use(hook_input: HookInput) -> HookResult:
+            if hook_input.tool_result is None:
+                return HookResult(allow=True)
+            redacted = _redact(hook_input.tool_result, secret_compiled)
```
DEENUU1, line 736: 🟡 **Redaction coerces structured tool results to `str`.** When a secret matches, `_post_tool_use` returns `modified_result=redacted` (a string), and `after_tool_execute` assigns it as the new result — but `read_file` is typed `-> Any` and some tools return non-string objects. For those, a secret hit silently changes the result *type* the model/downstream sees (object → str), while a non-match leaves the original object intact, so behavior is inconsistent depending on content. The comparison `redacted == hook_input.tool_result` is also against the already-`str()`-ified copy, not the real result. Low impact for the current string-returning tools, but worth gating redaction to cases where `result` is actually a `str` (or re-wrapping).

DEENUU1 (commented):
## Code review — PR #127 (security hook preset)

Reviewed `pydantic_deep/capabilities/hooks.py` in full, confirmed the tool names/arg names against `pydantic_ai_backends` console toolset (`path`/`command` — the handler wires up correctly, no fail-open there), and empirically exercised the regexes. The wiring is solid; the findings below are about the **regex coverage of a feature that markets itself as blocking these exact patterns**. The docs' "Limits" section already disclaims symlink/eval/env-var/novel-format bypasses, so I've excluded those — the items below are *not* covered by that disclaimer.

Severity: 🟠 medium · 🟡 low. Inline comments below.

### Verified empirically
```
BLOCK 'rm -rf /'              ALLOW 'rm -rf /*'
ALLOW 'rm -rf /home'          ALLOW 'rm --recursive --force /'
ALLOW 'rm -fr ~'
REDACT 'sk-abc...(old)'       LEAK 'sk-proj-...(current default)'
```

### Extra note (already disclaimed, mentioning for completeness)
The `read_file` sensitive-path blocklist is trivially sidestepped via `execute("cat /etc/shadow")`, since `execute` is only gated against the destructive-command patterns, not the read-path list. The docs' "not a sandbox / indirect file access" disclaimer covers this, so I'm not filing it as a finding — but if the read-path list is meant to be meaningful, consider also scanning `execute` commands for reads of those same paths.

PR #133, opened 2026-06-02
docs: render Material icons in examples overview

NumaCornec:
## Problem

On the [Examples overview page](https://vstorm-co.github.io/pydantic-deepagents/examples/#example-overview), the Material icon shortcodes (e.g. `:material-folder:`) render as **literal text** instead of icons.

## Cause

The overview cards place the icon shortcodes inside raw HTML (`<div>`/`<h3>`) blocks. Python-Markdown does not process Markdown (including `pymdownx.emoji`) inside raw block-level HTML unless the element opts in via the `md_in_html` `markdown` attribute. `md_in_html` is already enabled in `mkdocs.yml`, but the attribute was missing — so the shortcodes were emitted verbatim.

## Fix

Add the `markdown` attribute so the icon shortcodes get processed:
- `markdown` on the wrapping `.feature-grid` and each `.feature-card` div
- `markdown="span"` on each `<h3>`

Only `docs/examples/index.md` changes.

## Verification

`mkdocs build` on the changed page: 0 literal `:material-*:` shortcodes remain, and all 16 icons now render as inline `<span class="twemoji"><svg>…` SVGs.

### DIFF
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -16,100 +16,100 @@ uv run python examples/<example_name>.py
 
 ## Example Overview
 
-<div class="feature-grid">
+<div class="feature-grid" markdown>
 
-<div class="feature-card">
-<h3>:material-rocket-launch: Basic Usage</h3>
+<div class="feature-card" markdown>
+<h3 markdown="span">:material-rocket-launch: Basic Usage</h3>
 <p>Getting started with pydantic-deep. Create agents, use todos, work with files.</p>
 <a href="basic-usage/">View Example →</a>
 </div>
 
-<div class="feature-card">
-<h3>:material-folder: Filesystem</h3>
+<div class="feature-card" markdown>
+<h3 markdown="span">:material-folder: Filesystem</h3>
 <p>Real filesystem operations with FilesystemBackend.</p>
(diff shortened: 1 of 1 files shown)


NumaCornec:
Re: the Copilot suggestion to use `markdown="1"` instead of the bare `markdown` attribute, the bare form is correct here, and I verified it before opening the PR.

- Python-Markdown's `md_in_html` enables parsing for any `markdown` attribute that isn't `0`/`off`; a bare boolean `markdown` (empty value) is treated as enabled. `markdown="1"` and bare `markdown` behave identically.
- The bare form is also the idiom used in [Material for MkDocs' own grid/card docs](https://squidfunk.github.io/mkdocs-material/reference/grids/), so it matches the project's existing conventions.
- Verified with a full `mkdocs build` on this branch: the rendered `examples/index.html` contains **0** literal `:material-*:` shortcodes and **16** inline `<span class="twemoji"><svg>…` icons, one per card.

Happy to switch to `markdown="1"` if you'd prefer the explicit form for consistency, but functionally it's a no-op. (Note: the bot flagged it couldn't run its full agentic suite, so this wasn't build-verified on its side.)

PR #176, opened 2026-07-21
feat(cli): support OpenAI-compatible local endpoints

### LINKED ISSUE #175: Add llama.cpp support
### Problem Statement

Hello,

I don't want to use ollama because that forces me to use original models and i don't like them. 

Please add llama.cpp support it's so much better and faster

### Proposed Solution

With ollama the model has to get chat template feature or pydantic-deep doesn't work

### Alternatives Considered

_No response_

### Additional Context

_No response_

OchnikBartek:
### Summary

Let the CLI target any OpenAI-compatible local endpoint (llama.cpp, LM Studio, vLLM, text-generation-webui) by supplying a `base_url`. Previously only Ollama worked from the CLI, because it has a provider prefix pydantic-ai recognizes; other local servers need an `OpenAIChatModel` built with a custom `base_url`, which a plain model string couldn't carry.

### Business Context

Requested in #175. The author wants llama.cpp specifically (explicitly not Ollama). llama.cpp's `llama-server` already speaks the OpenAI wire format, so no new engine is needed — the only gap was a way to set the endpoint URL from the CLI.

### Changes

- **config** — add `base_url` and `local_api_key` fields to `CliConfig`.
- **providers** — add an `openai-compatible` provider entry and an `OPENAI_COMPATIBLE_PREFIX` sentinel.
- **agent** — `_resolve_openai_compatible_model()` turns an `openai-compatible:<name>` model string into an `OpenAIChatModel` pointed at `config.base_url` (noop API key when unset); wired into `create_cli_agent` so only prefixed models take the branch, everything else is untouched.
- **onboarding** — new `LocalEndpointModal` (base URL + model name + optional key); keyless-local providers now report as ready in the picker.
- **commands** — `/provider` routes the OpenAI-compatible choice to the new modal.
- **docs** — "Local models" section in `cli/settings.md` plus a tool-calling caveat.

The sentinel prefix keeps `base_url` from interfering with normal provider switching: it's only consulted when the active model is `openai-compatible:…`, and the model string round-trips through the existing config persistence unchanged.

### Verification

`tests/test_cli_local_models.py` (15 tests) covers the resolver (happy path, missing-`base_url` guard, default name), the `create_cli_agent` conversion branch, the `LocalEndpointModal` (connect / blank-name default / blank-URL warning / cancel), the `/provider` routing, and the onboarding status.

```
$ uv run pytest tests/test_cli_local_models.py
collected 15 items
tests/test_cli_local_models.py ...............                           [100%]
============================== 15 passed in 1.66s ==============================
```

Broader CLI/TUI suite (`-k "cli or tui or provider or model_picker or onboard or config"`): **607 passed**, no regressions.

Config written by the `/provider` → OpenAI-compatible flow:

```toml
model = "openai-compatible:qwen2.5"
base_url = "http://localhost:8080/v1"
local_api_key = ""   # most local servers ignore it
```

> Note: dev tooling (`ruff`, `pyright`, `mypy`) wasn't installed in the local env, so lint/typecheck ran via pre-commit only after `make install`. CI covers these gates.

### Linked Issues

Closes #175

### DIFF
--- a/apps/cli/agent.py
+++ b/apps/cli/agent.py
@@ -2,14 +2,15 @@
 
 from __future__ import annotations
 
+import os
 from collections.abc import Callable
 from pathlib import Path
-from typing import Any, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
 from pydantic_ai_backends import LocalBackend
 
 from apps.cli.config import CliConfig, load_config
-from apps.cli.providers import OPENAI_COMPATIBLE_PREFIX
+from apps.cli.providers import OPENAI_COMPATIBLE_API_KEY_ENV, OPENAI_COMPATIBLE_PREFIX
 from apps.cli.reminder import _build_reminder_config
 from pydantic_deep.agent import create_deep_agent
 from pydantic_deep.deps import DeepAgentDeps
@@ -17,6 +18,9 @@ from pydantic_deep.features.forking.capability import LiveForkCapability
 from pydantic_deep.features.hooks import Hook, HookEvent, HookInput, HookResult
 from pydantic_deep.features.message_queue import MessageQueue
 from pydantic_deep.prompts import build_system_prompt
+
+if TYPE_CHECKING:
+    from pydantic_ai.models.openai import OpenAIChatModel
 
 
 def _detect_fork_test_command(backend: Any) -> str | None:
@@ -110,13 +114,15 @@ def _make_shell_allow_list_hook(allow_list: list[str]) -> Hook:
     )
 
 
-def _resolve_openai_compatible_model(model_str: str, config: CliConfig) -> Any:
+def _resolve_openai_compatible_model(model_str: str, config: CliConfig) -> OpenAIChatModel:
     """Turn an `openai-compatible:<name>` model string into an `OpenAIChatModel`.
 
     Local servers (llama.cpp, LM Studio, vLLM, …) speak the OpenAI wire format but
     need a `base_url`, which a plain model string can't carry — so the CLI stores
-    the URL in `config.base_url` and marks the model with a sentinel prefix. Most
-    local servers ignore the API key, so a noop stand-in is used when none is set.
+    the URL in `config.base_url` and marks the model with a sentinel prefix. The
+    API key comes from the keystore (`OPENAI_COMPATIBLE_API_KEY`), never from
+    `config.toml`; most local servers ignore it, so a noop stand-in is used when
+    none is set.
     """
     from pydantic_ai.models.openai import OpenAIChatModel
     from pydantic_ai.providers.openai import OpenAIProvider
@@ -127,7 +133,8 @@ def _resolve_openai_compatible_model(model_str: str, config: CliConfig) -> Any:
             "Run /provider, pick OpenAI-compatible, and enter the server URL."
         )
     name = model_str[len(OPENAI_COMPATIBLE_PREFIX) :] or "local-model"
-    provider = OpenAIProvider(base_url=config.base_url, api_key=config.local_api_key or "sk-noop")
+    api_key = os.environ.get(OPENAI_COMPATIBLE_API_KEY_ENV) or "sk-noop"
+    provider = OpenAIProvider(base_url=config.base_url, api_key=api_key)
     return OpenAIChatModel(name, provider=provider)
 
 
@@ -468,7 +475,7 @@ def create_cli_agent(  # noqa: C901
         except Exception:
             mcp_servers = []
 
-    model_for_agent: Any = effective_model
+    model_for_agent: str | OpenAIChatModel = effective_model
     if isinstance(effective_model, str) and effective_model.startswith(OPENAI_COMPATIBLE_PREFIX):
         model_for_agent = _resolve_openai_compatible_model(effective_model, config)
 
@@ -542,10 +549,10 @@ def create_cli_agent(  # noqa: C901
             reminder_mode,
             config,
             on_reminder,
-            # Inherit the runtime model, not config.model (the TOML default,
-            # which may be a different provider than `-m` — e.g. Anthropic while
-            # running on Vertex, which then crashes with no ANTHROPIC_API_KEY).
-            reminder_model or config.reminder_model or effective_model,
+            # Inherit the resolved runtime model, not config.model (a different
+            # provider than `-m` would crash on a missing key; the raw
+            # `openai-compatible:...` string would break the LLM reminder generator).
+            reminder_model or config.reminder_model or model_for_agent,
         ),
     )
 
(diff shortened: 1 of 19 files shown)


apps/cli/agent.py
```diff
@@ -446,8 +468,12 @@ def _on_mcp_degraded(name: str, _reason: str) -> None:
         except Exception:
             mcp_servers = []
 
+    model_for_agent: Any = effective_model
+    if isinstance(effective_model, str) and effective_model.startswith(OPENAI_COMPATIBLE_PREFIX):
+        model_for_agent = _resolve_openai_compatible_model(effective_model, config)
```
DEENUU1: The conversion covers `model=`, but a few lines down `_build_reminder_config(...)` still receives `effective_model` — the raw `openai-compatible:...` string. The default `reminder_mode` is `"llm"`, so for exactly the users this PR targets, `LLMReminderGenerator` will fail to infer the model on every reminder and silently fall back to the zero-cost generator (with only a log warning).

I'd wire the resolved model through — hoist this block above the `create_deep_agent` call it already sits next to, pass `model_for_agent` as the last argument to `_build_reminder_config`, and widen `LLMReminderGenerator.model` to `str | Model` in `pydantic_deep/features/periodic_reminder/capability.py` (`Agent(model=...)` accepts both). That also keeps a local-only user's reminder traffic local, which is likely why they picked llama.cpp in the first place. Fine as a follow-up if you'd rather keep this PR tight, but then let's open an issue so it doesn't get lost.

apps/cli/agent.py
```diff
@@ -109,6 +110,27 @@ async def _check_command(hook_input: HookInput) -> HookResult:
     )
 
 
+def _resolve_openai_compatible_model(model_str: str, config: CliConfig) -> Any:
```
DEENUU1: nit: the return type can be precise — the function always returns an `OpenAIChatModel`. A `TYPE_CHECKING` import keeps the runtime import deferred:

```python
if TYPE_CHECKING:
    from pydantic_ai.models.openai import OpenAIChatModel


def _resolve_openai_compatible_model(model_str: str, config: CliConfig) -> OpenAIChatModel:
```

Same for `model_for_agent: Any` below — `str | OpenAIChatModel` says what it actually is.

apps/cli/providers.py
```diff
@@ -60,7 +60,18 @@ class ProviderInfo:
...
+        "OpenAI-compatible (local: llama.cpp / LM Studio / vLLM)",
+        "",
+        "",
+        "openai-compatible:local-model",
+    ),
 )
 
+#: Model-string prefix marking an OpenAI-compatible local endpoint. The part
+#: after the prefix is the model name; the endpoint URL lives in `config.base_url`.
+OPENAI_COMPATIBLE_PREFIX = "openai-compatible:"
```
DEENUU1, line 74: The sentinel-prefix design is the right call — the model string round-trips through config persistence and model history untouched, and nothing outside the prefixed branch ever consults `base_url`. The doc comment spelling that out saves the next reader a lot of digging.

apps/cli/screens/onboarding.py
```diff
@@ -175,3 +179,79 @@ def on_input_submitted(self, _event: Input.Submitted) -> None:
...
+
+        base_url = self.query_one("#local-url", Input).value.strip()
+        if not base_url:
+            self.app.notify("Please enter a base URL", severity="warning")
+            return
+        name = self.query_one("#local-model", Input).value.strip() or "local-model"
+        api_key = self.query_one("#local-key", Input).value.strip()
+
+        set_config_value(DEFAULT_CONFIG_PATH, "base_url", base_url)
+        set_config_value(DEFAULT_CONFIG_PATH, "local_api_key", api_key)
```
DEENUU1: Every other key in the CLI goes through the keystore (`save_key` → global `keys.toml` + `os.environ`); this one lands in the project's `config.toml`, which lives inside the working tree and is exactly the kind of file that ends up committed. Most local servers ignore the key, sure — but the moment someone points this at a remote OpenAI-compatible proxy (LM Studio remote, a vLLM deployment with `--api-key`), it's a real secret in a plain project file.

I'd route it through the keystore instead:

```python
if api_key:
    from apps.cli.keystore import save_key

    save_key("OPENAI_COMPATIBLE_API_KEY", api_key)
```

and have `_resolve_openai_compatible_model` read `os.environ.get("OPENAI_COMPATIBLE_API_KEY")` instead of `config.local_api_key`. That drops the `local_api_key` config field entirely — one less place a secret can live.

apps/cli/screens/onboarding.py
```diff
@@ -175,3 +179,79 @@ def on_input_submitted(self, _event: Input.Submitted) -> None:
...
+        self.dismiss(f"{OPENAI_COMPATIBLE_PREFIX}{name}")
+
+    def on_button_pressed(self, event: Button.Pressed) -> None:
+        if event.button.id == "btn-connect":
+            self._save_and_dismiss()
+        elif event.button.id == "btn-cancel":
+            self.dismiss(None)
+
+    def action_cancel(self) -> None:
+        self.dismiss(None)
```
DEENUU1, line 267: nit: `ApiKeyModal` submits on Enter; this modal makes you Tab down to the Connect button. Blank fields already have sensible defaults and a blank URL warns, so Enter-to-submit is safe from any field:

```suggestion
    def on_input_submitted(self, _event: Input.Submitted) -> None:
        self._save_and_dismiss()

    def action_cancel(self) -> None:
        self.dismiss(None)
```

DEENUU1 (commented):
Nice work — this closes #175 the right way: no new engine, just a way to carry `base_url`, and the sentinel prefix keeps it out of normal provider switching entirely. The test suite is genuinely good (the `create_deep_agent` spy that short-circuits with `_StopBuild`, and driving the modal callbacks directly), and the docs caveat about chat/tool templates addresses exactly what the issue author hit with Ollama.

Two things worth a look, both explained inline: the LLM reminder generator still receives the raw `openai-compatible:...` string, and the API key lands in `config.toml` instead of the keystore.

One more thing that's outside this diff but worth knowing: `_pick_available_model` (apps/cli/app.py:404) replaces the current model with a cloud default whenever `reconfigure_agent()` runs with no explicit model (settings toggle via `_apply_live`, closing the MCP view) and the model's provider has no key. A user who deliberately picked llama.cpp but has, say, `ANTHROPIC_API_KEY` in their env gets silently switched to a cloud model after toggling any setting. Ollama has the same problem today, so I'm fine with a follow-up rather than blocking here — but keyless prefixes (`ollama:`, `openai-compatible:`) should count as "explicitly chosen, keep it" in that helper. Amusing detail: `"openai-compatible:...".startswith("openai")` accidentally matches the OpenAI provider check, so with `OPENAI_API_KEY` set the local model survives by pure luck.

PR #188, opened 2026-07-28
feat: support enterprise oauth transport config for hosted mcp servers

### LINKED ISSUE #183: Enterprise-ready OAuth transport configuration for hosted MCP servers
### Problem Statement

I’m trying to configure Atlassian Rovo MCP as a hosted OAuth MCP server through MCPServerConfig, while keeping the setup declarative and reusable across projects.

The current config path supports OAuth, but it does not appear to expose enough HTTP/OAuth transport configuration for enterprise environments. In our case, `MCPServerConfig(auth=MCPAuth(kind="oauth"))` failed behind a corporate proxy/TLS inspection environment with certificate verification errors. We also need browser OAuth consent to persist after the first successful authorization, ideally in a user-level token store, so new TUI sessions and different repos do not  repeatedly ask the user to authenticate.

 The lower-level stack can handle this: `FastMCP/Pydantic AI` works when we use `OAuth(..., token_storage=..., httpx_client_factory=...)` with `StreamableHttpTransport(..., httpx_client_factory=...)`. The limitation is that using those directly means we lose the simple config-based MCP pattern.

This likely affects other enterprise hosted MCP servers too, especially Atlassian/Slack/GitHub-style OAuth servers used behind corporate proxies or custom CA chains.

### Proposed Solution

We worked around this by bypassing MCPServerConfig and building the MCP toolset directly:

```
  oauth = OAuth(
      mcp_url=ATLASSIAN_MCP_URL,
      callback_port=3120,
      scopes=scopes,
      token_storage=DiskStore(...),
      httpx_client_factory=enterprise_httpx_factory,
  )

  transport = StreamableHttpTransport(
      ATLASSIAN_MCP_URL,
      auth=oauth,
      httpx_client_factory=enterprise_httpx_factory,
  )

  mcp = MCPToolset(transport, id="atlassian", init_timeout=180)
```

That works, but it is much less reproducible than the config-based pattern and requires application-specific code for what is fundamentally transport/auth configuration.

We also tried API-token/Basic auth for Atlassian, but our organization does not allow API-token access for MCP product calls. OAuth is the correct path.

### Alternatives Considered

Atlassian Rovo MCP recommends the OAuth endpoint:

https://mcp.atlassian.com/v1/mcp/authv2

Using the lower-level FastMCP/Pydantic AI path, Atlassian OAuth succeeded and exposed the expected Jira/Confluence tools. The original config-based path failed in our environment because the HTTP/OAuth client needed proxy and truststore configuration.

This seems like an enterprise-readiness improvement for hosted MCP servers generally, not an Atlassian-specific feature.

### Additional Context

_No response_

OchnikBartek:
## Summary

Hosted OAuth MCP servers (Figma, Atlassian) can now be configured for enterprise
environments declaratively: `MCPAuth` gains explicit `scopes` and a fixed
`callback_port`, and a new `http_client_factory` escape hatch threads a custom
`httpx.AsyncClient` (authenticated proxy, OS trust store, mTLS) through both the
MCP transport and every step of the OAuth flow. Cached OAuth tokens are
namespaced per scope set, so changing `scopes` forces a fresh authorization
instead of silently reusing a token minted under the old permissions.

## Added

- Added `scopes` and `callback_port` to `MCPAuth` in `pydantic_deep/mcp/config.py` —
  validated, round-tripped through `to_dict`/`from_dict`, forwarded to FastMCP's
  `OAuth(...)` in `pydantic_deep/mcp/registry.py`.
- Added `HttpClientFactory` type and `http_client_factory=` kwarg on `MCPRegistry`
  and `build_mcp_server`. The factory is adapted to the MCP SDK's factory protocol
  (`_adapt_http_client_factory`: fresh client per call, with the `headers`/`timeout`/`auth`
  kwargs FastMCP passes applied on top) and reaches both the OAuth object and the
  transport, so discovery, token exchange, refresh and regular traffic all go
  through it.
- Added `_scoped_token_storage`: wraps the OAuth token store in a
  `PrefixKeysWrapper` keyed by a hash of the sorted scope set.
- Added docs: "Hosted OAuth servers" and "Corporate proxies and custom CAs"
  sections in `docs/learn/web-and-mcp.md` (env vars first, factory as last
  resort), plus `HttpClientFactory` in `docs/api/mcp.md`.

## Changed

- A config built with a client factory now constructs its
  `StreamableHttpTransport`/`SSETransport` explicitly (chosen from
  `config.transport`, not inferred from the URL), because `MCPToolset(url, ...)`
  has no way to carry an `httpx_client_factory` to the transport it builds
  internally. Configs without a factory build exactly as before.
- `mcp_oauth_storage()` in `apps/cli/mcp_store.py` logs a warning on both
  failure paths (missing disk backend, store init error) instead of silently
  degrading to in-memory token storage.

## Testing

- Added 13 tests across `tests/test_mcp.py` and `tests/test_cli_mcp.py`:
  field round-trip and validation, `OAuth` kwargs forwarding, factory identity
  asserted on both the OAuth object and the transport, fresh-client-per-call
  adapter contract, scope-set namespacing (order-insensitive, pass-through
  without scopes), SSE and stdio paths, registry threading, CLI fallback warnings.
- Ran `make lint`, `make typecheck` (pyright) and `make typecheck-mypy` — clean.
- Ran `make test` — 2811 passed, 100% coverage.
- Ran `mkdocs build` — no warnings.

## Related Issue

Closes #183 

## Notes for Reviewers

- Setting `scopes` on a server with an existing cached token triggers a one-time
  browser re-auth: the store key gains a scope-hash prefix, so the old
  un-prefixed token is deliberately not found. Configs without `scopes` keep
  using their existing tokens unchanged.
- The factory contract requires a fresh `httpx.AsyncClient` per call — both the
  transport and each OAuth step close the client they receive, so a shared
  instance would be dead after the first request. This is documented on
  `HttpClientFactory` and in the docs example.
- Deliberately out of scope, per the issue thread: a serializable `tls_verify`
  config field and an `atlassian` builtin catalogue entry.
- Verified with unit tests only — no access to an Atlassian tenant or a
  Negotiate proxy; the reporter offered to test end-to-end from their side.

### DIFF
--- a/apps/cli/mcp_store.py
+++ b/apps/cli/mcp_store.py
@@ -10,6 +10,7 @@ secret resolver.
 from __future__ import annotations
 
 import json
+import logging
 import os
 from pathlib import Path
 
@@ -37,6 +38,8 @@ __all__ = [
     "import_claude_code_servers",
 ]
 
+logger = logging.getLogger(__name__)
+
 
 def mcp_config_path() -> Path:
     """Path to the per-project MCP config file."""
@@ -58,17 +61,31 @@ def mcp_oauth_storage() -> object | None:
     Tokens (e.g. for the hosted Figma server) are cached on disk under
     ``~/.pydantic-deep/mcp-oauth`` and keyed by server URL, so authorizing once
     (via ``/mcp`` test) works for the agent too and survives restarts. Returns
-    ``None`` if the disk store backend isn't installed (falls back to in-memory).
+    ``None`` — with a logged warning, since the resulting in-memory fallback
+    means a browser re-auth on every restart — if the disk store backend isn't
+    installed or the store directory can't be created.
     """
     try:
         from key_value.aio.stores.disk import DiskStore
     except Exception:
+        logger.warning(
+            "MCP OAuth tokens will not persist: the disk store backend is not "
+            "installed (pip install 'py-key-value-aio[disk]'); using in-memory "
+            "storage, so hosted servers re-authorize on every restart."
+        )
         return None
     path = Path.home() / ".pydantic-deep" / "mcp-oauth"
     try:
         path.mkdir(parents=True, exist_ok=True)
         return DiskStore(directory=str(path))
-    except Exception:
+    except Exception as exc:
+        logger.warning(
+            "MCP OAuth tokens will not persist: cannot initialise the disk store "
+            "at %s (%s); using in-memory storage, so hosted servers re-authorize "
+            "on every restart.",
+            path,
+            exc,
+        )
         return None
 
 
--- a/pydantic_deep/mcp/config.py
+++ b/pydantic_deep/mcp/config.py
@@ -60,6 +60,14 @@ class MCPAuth:
         client_name: OAuth client name advertised during dynamic client
             registration (``oauth`` only). Some servers (e.g. Figma's hosted MCP
             during beta) allowlist this; leave ``None`` to use the default.
+        scopes: OAuth scopes to request up front (``oauth`` only). Some hosted
+            servers (e.g. Atlassian's) require explicit scopes at registration
+            instead of granting a default set. Cached tokens are namespaced per
+            scope set, so changing this forces a fresh authorization rather
+            than silently reusing a token minted under the old scopes.
+        callback_port: Fixed localhost port for the OAuth redirect URI
+            (``oauth`` only). Needed by servers that validate a pre-registered
+            redirect URI; ``None`` lets the client pick a free port.
     """
 
     secret_key: str = ""
@@ -69,10 +77,14 @@ class MCPAuth:
     value_template: str = "Bearer {token}"
     instructions: str = ""
     client_name: str | None = None
+    scopes: list[str] = field(default_factory=list)
+    callback_port: int | None = None
 
     def __post_init__(self) -> None:
         if self.kind in _SECRET_AUTH_KINDS and not self.secret_key:
             raise MCPConfigError(f"{self.kind} MCP auth requires a non-empty secret_key")
+        if self.callback_port is not None and self.callback_port <= 0:
+            raise MCPConfigError(f"callback_port must be positive, got {self.callback_port}")
         # Fail fast on a template that would raise at `render_value` time —
         # extra placeholders (`{foo}`) or unbalanced literal braces (B13).
         try:
@@ -97,6 +109,8 @@ class MCPAuth:
             value_template=data.get("value_template", "Bearer {token}"),
             instructions=data.get("instructions", ""),
             client_name=data.get("client_name"),
+            scopes=list(data.get("scopes", [])),
+            callback_port=data.get("callback_port"),
         )
 
 
--- a/pydantic_deep/mcp/registry.py
+++ b/pydantic_deep/mcp/registry.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import hashlib
 import logging
 import os
 import re
@@ -23,7 +24,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, cast
 
-from pydantic_deep.mcp.config import _SECRET_AUTH_KINDS, MCPServerConfig
+from pydantic_deep.mcp.config import _SECRET_AUTH_KINDS, MCPAuth, MCPServerConfig
 from pydantic_deep.mcp.resources import create_mcp_resources_toolset
 
 
@@ -43,6 +44,7 @@ def _stdio_log_file(name: str) -> Path:
 
 
 if TYPE_CHECKING:
+    import httpx
     from pydantic_ai._run_context import RunContext
     from pydantic_ai.toolsets import AbstractToolset
     from pydantic_ai.toolsets.abstract import ToolsetTool
@@ -51,6 +53,7 @@ logger = logging.getLogger("pydantic_deep.mcp")
 
 __all__ = [
     "SecretResolver",
+    "HttpClientFactory",
     "MCPRegistry",
     "MCPProbeResult",
     "MCPNotInstalledError",
@@ -62,6 +65,16 @@ __all__ = [
 
 SecretResolver = Callable[[str], "str | None"]
 """Resolves a secret key (e.g. ``"GITHUB_MCP_PAT"``) to its token, or ``None``."""
+
+HttpClientFactory = Callable[[MCPServerConfig], "httpx.AsyncClient"]
+"""Builds an ``httpx.AsyncClient`` for a server config (corporate proxy,
+custom CA / OS trust store, mTLS, authenticated proxy headers, …).
+
+Every HTTP-based connection goes through it — the MCP transport *and* each
+step of the OAuth flow (discovery, registration, token exchange, refresh) —
+and each call site closes the client it receives, so the factory must return
+a **fresh client per call**, never a shared instance.
+"""
 
 MCP_INSTALL_HINT = (
     "MCP support requires the optional 'mcp' extra: "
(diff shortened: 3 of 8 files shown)


pydantic_deep/mcp/registry.py
```diff
@@ -111,11 +124,104 @@ def auth_satisfied(config: MCPServerConfig, resolver: SecretResolver | None = No
...
+        auth: httpx.Auth | None = None,
+    ) -> httpx.AsyncClient:
+        client = factory(config)
+        if headers:
+            client.headers.update(headers)
+        if timeout is not None:
+            client.timeout = timeout
+        if auth is not None:
+            client.auth = auth
+        return client
```
DEENUU1, line 151: **Blocking.** This signature is one kwarg short of what FastMCP actually calls, so the whole feature breaks on `http` (the default, and what Atlassian uses).

`StreamableHttpTransport.connect_session` does:

```python
http_client = self.httpx_client_factory(
    headers=headers,
    auth=self.auth,
    follow_redirects=True,  # type: ignore[call-arg]
    **({"timeout": timeout} if timeout else {}),
)
```

`follow_redirects` isn't in the SDK's `McpHttpClientFactory` protocol — FastMCP passes it anyway with a `type: ignore`. pydantic-ai hit the same thing and their `_make_httpx_client_factory` carries a comment about it. I ran this on your branch against a dead port:

```
no factory : Client failed to connect: All connection attempts failed
w/ factory : Client failed to connect: _adapt_http_client_factory.<locals>.make_client() got an unexpected keyword argument 'follow_redirects'
```

And with `auth=MCPAuth(kind="oauth")` it's the same `TypeError`. So today, any HTTP server built with a factory fails at connect — and because `make_resilient` turns that into "server unavailable", the user just sees zero tools with no clue why.

Accepting it also fixes a second thing worth having: a plain `httpx.AsyncClient()` from a user factory has `follow_redirects=False`, while the MCP default path (`create_mcp_http_client`) always sets it to `True`. Applying the kwarg keeps the two paths behaving the same.

```suggestion
    def make_client(
        headers: dict[str, str] | None = None,
        timeout: httpx.Timeout | None = None,
        auth: httpx.Auth | None = None,
        # Not in the MCP SDK's `McpHttpClientFactory` protocol, but FastMCP's
        # `StreamableHttpTransport` passes it anyway — omitting it is a
        # `TypeError` on the first connect.
        follow_redirects: bool = True,
    ) -> httpx.AsyncClient:
        client = factory(config)
        if headers:
            client.headers.update(headers)
        if timeout is not None:
            client.timeout = timeout
        if auth is not None:
            client.auth = auth
        client.follow_redirects = follow_redirects
        return client
```

I applied exactly this locally and re-ran the probe — `http`, `sse` and `http`+oauth all get to "All connection attempts failed", i.e. they reach the network instead of dying in the adapter.

tests/test_mcp.py
```diff
@@ -171,6 +171,203 @@ def test_registry_threads_oauth_storage(tmp_path: Path) -> None:
...
+    def factory(config: MCPServerConfig) -> httpx.AsyncClient:
+        assert config is cfg
+        client = httpx.AsyncClient(headers={"Proxy-Authorization": "Negotiate abc"})
+        made.append(client)
+        return client
+
+    adapted = _adapt_http_client_factory(factory, cfg)
+    auth = httpx.Auth()
+    first = adapted()  # the OAuth flow calls with no arguments
+    second = adapted(headers={"X-H": "1"}, timeout=httpx.Timeout(7.0), auth=auth)
```
DEENUU1, line 307: This is the test that should have caught the bug above, and it doesn't — because it calls the adapter with the kwargs the *protocol* declares rather than the kwargs FastMCP actually passes. That's why 13 tests, 100% coverage and green CI all agreed the feature worked.

Worth pinning the real call shape so a future FastMCP change breaks a test instead of a user's connection:

```python
    # The exact kwargs FastMCP's StreamableHttpTransport.connect_session passes —
    # `follow_redirects` is not in the MCP SDK's declared factory protocol.
    third = adapted(headers={}, auth=auth, follow_redirects=True)
    assert third.follow_redirects is True
    made.append(third)
```

Even better, if it's cheap here: one test that builds an http server with a factory and probes it against a closed port, asserting the error isn't a `TypeError`. That covers the whole wiring rather than the adapter in isolation.

pydantic_deep/mcp/registry.py
```diff
@@ -111,11 +124,104 @@ def auth_satisfied(config: MCPServerConfig, resolver: SecretResolver | None = No
...
+    """Build the FastMCP HTTP/SSE transport explicitly.
+
+    ``MCPToolset(url, ...)`` offers no way to hand an ``httpx_client_factory``
+    to the transport it builds internally, so a config with a client factory
+    constructs the transport itself — chosen from ``config.transport`` rather
+    than inferred from the URL — and passes it to the toolset.
+    """
+    from fastmcp.client.transports import SSETransport, StreamableHttpTransport
+
+    transport_cls = SSETransport if config.transport == "sse" else StreamableHttpTransport
```
DEENUU1, line 213: Choosing the class from `config.transport` is the right call — but it only happens when a factory is passed, and that makes the same config mean two different things depending on an unrelated knob.

Without a factory we go through `MCPToolset(url, ...)`, which infers via FastMCP's `infer_transport_type_from_url` — it returns `"sse"` only when the path matches `/sse(/|?|&|$)`. So:

- `transport="sse"`, `url="https://x/events"` → StreamableHttp today, SSE the moment someone sets `http_client_factory` on the registry.
- `transport="http"`, `url="https://x/sse"` → SSE today, StreamableHttp with a factory.

Nobody will connect the switch to the factory when that bites. I'd rather the declared transport always win: build the transport explicitly in both branches and drop the `http_client_factory is not None` check (passing `httpx_client_factory=None` is fine — FastMCP treats it as absent). One code path, and `transport="sse"` means SSE.

If you'd rather not widen the diff, that's fair — but then please say in the `_build_http_transport` docstring that this path can disagree with the inferred one, so the next person doesn't have to rediscover it.

apps/cli/mcp_store.py
```diff
@@ -58,17 +61,31 @@ def mcp_oauth_storage() -> object | None:
...
     (via ``/mcp`` test) works for the agent too and survives restarts. Returns
-    ``None`` if the disk store backend isn't installed (falls back to in-memory).
+    ``None`` — with a logged warning, since the resulting in-memory fallback
+    means a browser re-auth on every restart — if the disk store backend isn't
+    installed or the store directory can't be created.
     """
     try:
         from key_value.aio.stores.disk import DiskStore
     except Exception:
+        logger.warning(
```
DEENUU1, line 71: Right instinct — the silent fallback was the confusing part — but `logging.getLogger(__name__)` is the wrong channel here, and it can make things worse than staying quiet.

`load_mcp_registry()` is called from inside running Textual modals (`apps/cli/modals/mcp_view.py:178`, `apps/cli/modals/info_view.py:97`). `apps.cli.mcp_store` isn't in `_NOISY_CONSOLE_LOGGERS`, so it propagates to root; root has no handler under the TUI, so `logging.lastResort` writes the record straight to stderr — over the live screen. That's exactly what `quiet_console_logging()` and the stdio log-file redirect exist to prevent. And the user who needs this message (repeated re-auth) still doesn't get anything actionable in the UI.

The house pattern for CLI-side warnings is the session logger — `propagate=False`, file handler, no screen corruption:

```python
from apps.cli.debug_log import get_logger

get_logger().warning(
    "MCP OAuth tokens will not persist: the disk store backend is not "
    "installed (pip install 'py-key-value-aio[disk]'); using in-memory "
    "storage, so hosted servers re-authorize on every restart."
)
```

Same for the second `logger.warning` below. If you want the user to actually see it, the `/mcp` view already has a place to surface it — but the file log is the minimum and the safe default.

pydantic_deep/mcp/config.py
```diff
@@ -69,10 +77,14 @@ class MCPAuth:
...
     instructions: str = ""
     client_name: str | None = None
+    scopes: list[str] = field(default_factory=list)
+    callback_port: int | None = None
 
     def __post_init__(self) -> None:
         if self.kind in _SECRET_AUTH_KINDS and not self.secret_key:
             raise MCPConfigError(f"{self.kind} MCP auth requires a non-empty secret_key")
+        if self.callback_port is not None and self.callback_port <= 0:
+            raise MCPConfigError(f"callback_port must be positive, got {self.callback_port}")
```
DEENUU1, line 87: nit: while you're validating, the upper bound is worth the same line. `callback_port=70000` passes here and then fails much later with a socket error, in the middle of a browser flow, where it's hard to trace back to the config.

```suggestion
        if self.callback_port is not None and not 0 < self.callback_port <= 65535:
            raise MCPConfigError(
                f"callback_port must be between 1 and 65535, got {self.callback_port}"
            )
```

tests/test_mcp.py
```diff
@@ -171,6 +171,203 @@ def test_registry_threads_oauth_storage(tmp_path: Path) -> None:
...
+    cfg = MCPServerConfig(
+        name="atlassian", transport="http", url="https://x/mcp", auth=MCPAuth(kind="oauth")
+    )
+    ts = build_mcp_server(cfg, lambda k: None, http_client_factory=factory)
+    transport = _transport(ts)
+    oauth = transport.auth
+    # Configuring only the transport would just move the failure from
+    # `initialize` to the first token refresh — both sides need the factory.
+    assert transport.httpx_client_factory is not None
+    assert oauth.httpx_client_factory is transport.httpx_client_factory
```
DEENUU1, line 286: This assertion is the best line in the PR. Configuring only the transport moves the failure from `initialize` to the first token refresh, which is a horrible thing to debug months later, and this pins it as an identity check rather than a "not None". Same for `test_scoped_token_storage_namespaces_by_scope_set` asserting order-insensitivity and the no-scopes pass-through — that pair is what makes the cache-invalidation change safe for people with existing tokens.

DEENUU1 (changes requested):
Thanks for this — the shape is right and you clearly read the layers underneath before writing any code. Splitting `_build_oauth` / `_build_http_transport` / `_adapt_http_client_factory` out keeps `_build_raw_mcp_toolset` readable, the scope-set namespacing is a neat fix for a cache bug that would otherwise be nearly undiagnosable, and the test that asserts the OAuth object and the transport share the *same* factory object is exactly the trap I was worried about. Docs are good too — env vars first, factory last.

One blocking problem: the factory never actually works over `http`. FastMCP's `StreamableHttpTransport.connect_session` calls the factory with a `follow_redirects=True` kwarg that the MCP SDK's `McpHttpClientFactory` protocol doesn't declare, so `make_client` raises `TypeError` on the first connect. I reproduced it against a dead port on this branch:

```
no factory : Client failed to connect: All connection attempts failed
w/ factory : Client failed to connect: _adapt_http_client_factory.<locals>.make_client() got an unexpected keyword argument 'follow_redirects'
```

Same for the OAuth+factory path, which is the Atlassian case this PR exists for. The `sse` path is fine (the SDK's `sse_client` sticks to the declared protocol). Details and a one-line fix inline — I verified the patched version reaches the network on all three paths.

The rest is smaller: one behaviour divergence in transport selection, one about where the new CLI warning lands under the TUI, and a nit.

pydantic_deep/mcp/registry.py
```diff
@@ -111,11 +124,104 @@ def auth_satisfied(config: MCPServerConfig, resolver: SecretResolver | None = No
...
+    client_factory: Callable[..., httpx.AsyncClient] | None,
+) -> Any:
+    """Construct the FastMCP ``OAuth`` object for an ``oauth``-kind config."""
+    from fastmcp.client.auth.oauth import OAuth
+
+    oauth_kwargs: dict[str, Any] = {"mcp_url": cast(str, config.url)}
+    if oauth_token_storage is not None:
+        storage = oauth_token_storage
+        if auth.scopes:
+            storage = _scoped_token_storage(storage, auth.scopes)
```
DEENUU1, line 184: The namespace is keyed on `scopes` alone, but the thing being namespaced is the *whole* OAuth cache, and the cached dynamic client registration depends on more than scopes.

`PrefixKeysWrapper` prefixes every key, so this also namespaces the DCR result (FastMCP stores it via `key=_get_client_info_cache_key()` in the `mcp-oauth-client-info` collection). That's good — but the registration is built from `client_metadata`, which includes `redirect_uris=[http://localhost:{callback_port}/callback]` and `client_name`, not just `scope`. So `callback_port` and `client_name` change the registration identity without changing the cache slot.

What that means for someone who already authorized a server and then adds `callback_port` on its own (which the new docs present as an independent knob, right next to `scopes`):

- `auth.scopes` is empty → no prefix → the old cached `client_info` is found
- the SDK only registers when there is no cached client info (`mcp/client/auth/oauth2.py`: `if not self.context.client_info:`)
- so the authorization request goes out with the **old** `client_id` and the **new** `redirect_uri` (`str(self.context.client_metadata.redirect_uris[0])`)
- the server sees a redirect URI it never registered

Hedging honestly: this may self-heal. FastMCP's `redirect_handler` pre-flights the authorization URL and turns a **400** into `ClientNotFoundError`, which clears the store and retries with a fresh registration. RFC 6749 §4.1.2.1 says a server must not redirect on a bad `redirect_uri`, so a 400 is likely and many providers will recover after one extra round trip. But a provider that answers 200 with an HTML error page passes FastMCP's `status_code not in (200, 302, 303, 307, 308)` check, and the user just gets a browser tab on an error with no way out but deleting `~/.pydantic-deep/mcp-oauth` by hand. I verified the code path, not a live tenant — no Atlassian access here either.

Since the whole point of this function is "don't silently reuse credentials minted under different terms", I'd key it on everything that goes into `client_metadata` and drop the dependency on that recovery path:

```python
def _oauth_cache_namespace(auth: MCPAuth) -> str | None:
    """Cache namespace for an OAuth config's tokens *and* client registration.

    FastMCP keys both by server URL only, so a config change that alters the
    registered client — scopes, the redirect URI's port, the advertised client
    name — would otherwise keep reusing credentials minted under the old terms.
    ``None`` when there is nothing to namespace, so tokens cached before this
    existed keep working.
    """
    identity = (
        ",".join(sorted(set(auth.scopes))),
        str(auth.callback_port or ""),
        auth.client_name or "",
    )
    if not any(identity):
        return None
    digest = hashlib.sha256("|".join(identity).encode()).hexdigest()[:16]
    return f"oauth-{digest}"
```

and here:

```python
    if oauth_token_storage is not None:
        storage = oauth_token_storage
        namespace = _oauth_cache_namespace(auth)
        if namespace is not None:
            storage = PrefixKeysWrapper(key_value=storage, prefix=namespace)
        oauth_kwargs["token_storage"] = storage
```

Two notes on the blast radius, both checked:

- No builtin sets `client_name` (the Figma entry is `MCPAuth(kind="oauth", instructions=...)` only), so no shipped config gains a prefix from this. A hand-written config with `client_name` set would re-auth once — the same one-time cost you already documented for `scopes` in *Notes for Reviewers*, and arguably a fix: today, changing `client_name` silently has no effect at all, because the old registered name is reused.
- `scopes`-only configs would get a different digest than the current one, but nothing is released yet, so no real token is affected.

nit folded in: `sorted(set(...))` rather than `sorted(...)`. `test_scoped_token_storage_namespaces_by_scope_set` says the namespace is a function of the scope *set*, and with duplicates it currently isn't — `["read:jira", "read:jira"]` and `["read:jira"]` request identical scopes but land in different slots.

DEENUU1 (commented):
Second pass over the parts I skimmed the first time. One thing worth changing while you're in here, plus a nit folded into the same spot — nothing else new. I re-checked the transport/timeout wiring (pydantic-ai does pass `timeout=read_timeout` down to the session, so the adapter's `client.timeout` covers it), the `py-key-value-aio>=0.4.5` floor (has `PrefixKeysWrapper`), the docs heading nesting, and the CLI add-server flow (create-only, so no field-dropping on edit) — all fine.

PR #190, opened 2026-07-31
fix(teams): repair subagent wiring so members actually run (0.3.40)

DEENUU1:
Agent teams delegate execution to the subagent engine, and never actually ran a
member. Four defects compounded, each of them silent — the tools reported success
and the team produced nothing, so the lead saw work perpetually "running".

> **Stacked on #186** (`0.3.39`), which this branch was built on. Base is `0.3.39`,
> not `main`.
>
> **Requires `subagents-pydantic-ai >= 0.2.12`** — vstorm-co/pydantic-ai-subagents#58.
> `message_teammate` calls `SubAgentToolset.answer_task` / `steer_task`, and the
> registry fix reads `SubAgentToolset.registry`; neither exists in 0.2.11. The
> dependency floor is deliberately **not** raised here, because 0.2.12 is not on
> PyPI yet and the lock could not resolve. Merge after that release, then let
> renovate raise the floor (cf. #189).

## The four defects

**1. Teams registered members in a registry the subagent engine never reads.**

With `subagent_registry=None` — the `create_deep_agent` default — the team block in
`agent.py` built a fresh `DynamicAgentRegistry()`. `spawn_team` registered members
into it; the subagent `task` tool resolves names against its own. Every
`assign_task` came back `Error: Unknown subagent`. Teams plus subagents was broken
in the default configuration:

```
registered in team registry: ['coder']
visible to subagent toolset: []
assign: Task assigned to 'coder'. Agent running in background.
Error: Unknown subagent 'coder'. Available:
```

Teams now share `subagent_toolset.registry`.

**2. A refused delegation was recorded as a started task.**

The subagent `task` tool reports an unknown subagent, or a busy chat trace, as an
error *string* rather than raising — so `assign_task`'s `except` never fired. It set
the member to `running` and appended the error text to an "Agent running in
background" message, and the member sat at `running` forever. A started task is now
identified by its `Task ID:` line, and a refusal marks the member `failed` with the
reason.

**3. `message_teammate` delivered nothing.**

It wrote to `TeamMessageBus`, which nothing running reads: a member is a background
subagent and only sees what the subagent engine hands it, and
`TeamMessageBus.receive()` is called from tests only. The tool reported "Message
sent to 'alice'" and the message sat on a queue forever.

It now routes through the engine — `answer_task` when the member is blocked in
`ask_parent`, `steer_task` while it runs — and says plainly when there is nothing to
deliver into. End to end, a member can now ask a question and get an answer:

```
msg: Answer delivered to 'coder', which was waiting for one.
- coder: completed
```

**4. Shared todos never closed.**

A finished member left its task `in_progress` on the shared list, so the lead read
completed work as outstanding. `AgentTeam.sync_member` completes the todo when a
member completes, and *releases the claim* when one fails, so the lead can reassign
it instead of watching a dead member hold it.

## Changes

- `agent.py`: the team registry falls back to `subagent_toolset.registry`, and the
  toolset itself is passed through as `subagent_toolset`.
- `features/teams/toolset.py`: `_extract_task_id` decides started-vs-refused;
  `message_teammate` delivers through the subagent engine; `check_teammates` syncs
  through `AgentTeam.sync_member`.
- `features/teams/primitives.py`: `TeamMemberHandle.todo_id` links a member to the
  item it claimed; `AgentTeam.sync_member` closes or releases that todo.
- `create_team_toolset` gains `subagent_toolset`. The existing `registry` /
  `task_fn` / `task_manager` arguments are unchanged.

## Tests

`TestTeamSubagentWiring` — 7 tests, each failing on its own defect (verified by
reverting the fixes one at a time). `test_default_agent_shares_the_subagent_registry`
drives the toolsets `create_deep_agent` actually builds, so it covers the wiring
rather than the test's own setup.

`TestCreateTeamToolset::test_message_teammate` asserted the `"Message sent"` that
defect 3 shows was a lie. It now asserts the honest outcome and is renamed to say
what it checks.

Full suite: **2781 passed**.

## Not changed

Team members get `ask_parent` with the subagent default 300-second timeout, so a
member that asks while the lead is not polling blocks for five minutes. That is a
policy decision rather than a bug — `subagents-pydantic-ai` 0.2.12 exposes
`ask_timeout_seconds` if we want teams to set something shorter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,55 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.40] - 2026-08-01
+
+Agent teams were wired to the subagent execution engine but never actually ran a
+member. Four defects compounded, each of them silent: the tools reported success
+and the team produced nothing.
+
+Requires `subagents-pydantic-ai >= 0.2.12`, which adds the programmatic
+`answer_task` / `steer_task` surface and exposes `SubAgentToolset.registry`.
+
+### Fixed
+
+- **Teams registered members in a registry the subagent engine never reads.** With
+  `subagent_registry=None` — the `create_deep_agent` default — the team block built
+  a fresh `DynamicAgentRegistry()`, so `spawn_team` put its members somewhere the
+  subagent `task` tool does not look and every `assign_task` came back
+  `Error: Unknown subagent`. Teams plus subagents was broken in the default
+  configuration. Teams now share `subagent_toolset.registry`.
+- **A refused delegation was recorded as a started task.** The subagent `task` tool
+  reports an unknown subagent or a busy chat trace as an error string rather than
+  raising, so `assign_task`'s `except` never fired: it set the member to `running`
+  and appended the error to an "Agent running in background" message. The member
+  then sat at `running` forever. A started task is now identified by its `Task ID:`
+  line, and a refusal marks the member `failed` with the reason.
+- **`message_teammate` delivered nothing.** It wrote to `TeamMessageBus`, which no
+  running member reads — a member is a background subagent and only sees what the
+  subagent engine hands it, and `TeamMessageBus.receive()` is called from tests
+  only. It now routes through the engine: `answer_task` when the member is blocked
+  in `ask_parent`, `steer_task` while it runs, and it says plainly when there is
+  nothing to deliver into instead of reporting "Message sent".
+- **A finished member left its shared task `in_progress` forever**, so the lead read
+  completed work as outstanding. `AgentTeam.sync_member` completes the todo when a
+  member completes, and releases the claim when one fails so the lead can reassign
+  it rather than watching a dead member hold it.
+
+### Added
+
+- **`create_team_toolset(subagent_toolset=...)`** — the subagent toolset backing
+  execution, which is what makes `message_teammate` able to deliver. The existing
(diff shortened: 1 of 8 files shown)


DEENUU1:
Marked draft — blocked on the `subagents-pydantic-ai` 0.2.12 release (vstorm-co/pydantic-ai-subagents#58).

**No CI runs on this PR.** `ci.yml` triggers only on `pull_request` targeting `main`, and this targets `0.3.39` (#186). Ran the CI jobs locally instead:

| Job | Result |
|---|---|
| ruff format / check | pass (292 files) |
| pytest | **2781 passed** |
| bandit | pass (exit 0, same as base) |
| mypy vs 0.2.12 | pass (169 files) |
| mypy vs installed 0.2.10 | 1 error, expected — see below |
| pyright vs installed 0.2.10 | 1 error, same cause |

The one remaining error is version skew, not a defect: `SubAgentToolset` is a function alias in 0.2.11 and a class from 0.2.12, so `subagent_toolset: SubAgentToolset | None` is only a valid annotation once the floor moves. Verified by resolving against 0.2.12, where it disappears.

`4943126` fixes the genuine fallout from 0.2.12's stricter types: `SubAgentConfig` now enforces `instructions` (three test literals omitted it), and `ToolsetFactory` returns a `Sequence` so elements are no longer `Any`.

**Merge order:** release subagents 0.2.12 → merge #186 → raise the dependency floor here → retarget this to `main` so CI actually runs → merge.

DEENUU1:
Auto-closed when the base branch `0.3.39` was deleted on merge of #186, and GitHub will not reopen a PR whose base is gone. Continued as #191 — same branch and commits, based on `main` so CI runs.

